PR #6059, opened 2013-08-15
New CSS Class for DCA / Field / Eval

Total-Reality:
https://contao.org/en/manual/3.0/data-container-arrays.html#arranging-fields
1. Please add a new css class like "subarea" with the following style in the theme: margin-top:15px;border-top:1px solid #ddd;padding-top:15px

Background: In my extension I want to show addImage and addGallery in one fieldset (legend). Without margin und a border it is too tight.
1. Please change the behavior for the eval attribute "style"
   Currently this has an effect to the input tag. It would be better if changes influence the encircling div tag.


Total-Reality:
"Why don't you make this two different palettes?"
You mean two different legends? Because it is confusing, I already have 20 legends or so in this extension...

PR #6289, opened 2013-10-09
fix for #5030 

fbuchinger:
remove wrong %s placeholders in spanish lang file that causes sprintf error in install.php


fbuchinger:
Please reopen the bug - The wrong placeholders are still present in contao 3.4

Toflar:
And translations must still be fixed on transifex.com.

PR #7034, opened 2014-05-23
Rich text editor file in own module folder

tsarma:
Please make it possible to place rte setting file in your own module folder instead of confined to /system/config folder.
Usage: 

``` php
 'eval' => array('rte'=>'tinyPath::/system/modules/myModule/assets/myRTE.php', 
```

I choose "tinyPath::" because inside class DataContainer.php your are looking for the occurrence of word 'tiny' in first 4 characters.

``` php
//line 445
if (isset($arrData['eval']['rte']) && strncmp($arrData['eval']['rte'], 'tiny', 4) === 0)
```

### DIFF
--- a/system/modules/core/classes/BackendTemplate.php
+++ b/system/modules/core/classes/BackendTemplate.php
@@ -83,7 +83,12 @@ class BackendTemplate extends \Template
 					$file = 'ace';
 				}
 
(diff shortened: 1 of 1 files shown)


leofeyer:
This will be solved in Contao 4 where we have made the RTE config files real templates (e.g. `ce_tinyMCE.html5`). These templates can be part of a third-party extension and they can easily be overwritten in the template editor.

PR #7137, opened 2014-06-30
.htaccess.default - Trailing slash fix + 1 bugfix

maan:
Trailing slash fix: In case no suffix is used at all, now there's an option to add a slash to the request URI. Until now, without an ending slash, you get page not found errors with Contao, which isn't very handy.

Added another fix extending some checking if we're dealing with a real file/directory, or not.

### DIFF
--- a/.htaccess.default
+++ b/.htaccess.default
@@ -224,6 +224,17 @@ FileETag None
   ##
 
   ##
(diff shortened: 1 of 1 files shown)


leofeyer:
I'm closing this ticket, because we are no longer supporting `.htaccess` files in Contao 4 and the `.htaccess.default` file will be removed.

PR #7165, opened 2014-07-11
Added universal polyfills for developers

Toflar:
We currently have the possibility to develop our JavaScripts either using jQuery or MooTools. Very often I prefer to use none of those libraries and rely on pure (Vanilla) JS only and I think I'm not the only developer doing so :-)

The problem I always have - and this is especially true when I'm developing extensions that can be used anywhere rather than just for a single customer project - are polyfills. Polyfills are great because they ensure you can write normal JavaScript for newer browsers and the polyfill will take care for older ones so my code just works fine everywhere.

However, we have the problem that everybody ships their own polyfills. Extension A ships the same polyfill as extension B and C because they all need it and don't know of each other. Of course this creates a useless overhead of data.
What makes it even worse is that a lot of those polyfills are loaded even if they are not needed at all because the browser already supports the needed feature.

So I wanted to create a universal solution and searched the internet for something that lazy loads polyfills. And I eventually found https://github.com/kbjr/polyfill.js :-)

Polyfill.js dynamically loads required polyfills only when they're needed and as soon as they're ready, executes your code.
I have implemented a very easy solution. Contao will load the core of Polyfill.js only if a developer wants it to do so and any expert can still disable it in the page layout.
## Usage

Let's say you as a developer need the `window.onhashchange` event. Relying on a polyfill is now as easy as that:

**config.php**

``` php
$GLOBALS['TL_JS_POLYFILLS'][] = 'hashchange';
```

**Your template code**

``` html
<script>
window.addEventListener('ContaoPolyfillsLoaded', function() {
    // You can now savely use window.onhashchange in your code
});
</script>
```
## What polyfills can I use?

Anything from that list: https://github.com/kbjr/polyfill.js#available-polyfills
## I need a different polyfill!

Contribute! You can do that easily, see: https://github.com/kbjr/polyfill.js#contributing-a-new-polyfill
And the good thing about it is that Contao does not need to change. As soon as it's merged it will be available via `polyfill.herokuapp.com` :)

### DIFF
--- a/system/modules/core/config/config.php
+++ b/system/modules/core/config/config.php
@@ -464,6 +464,10 @@ $GLOBALS['TL_ASSETS'] = array
 	'SLIMBOX'      => '1.8'
 );
 
+/**
+ * JavaScript Polyfills
+ */
+$GLOBALS['TL_JS_POLYFILLS'] = array();
 
 /**
  * Other global arrays
--- a/system/modules/core/dca/tl_layout.php
+++ b/system/modules/core/dca/tl_layout.php
@@ -99,7 +99,7 @@ $GLOBALS['TL_DCA']['tl_layout'] = array
 	'palettes' => array
 	(
 		'__selector__'                => array('rows', 'cols', 'addJQuery', 'addMooTools', 'static'),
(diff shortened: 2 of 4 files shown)


Toflar:
I think we all agree that polyfills should be handled as dependencies just as packages. I think we should discuss integration of https://github.com/puli for normal as well as web assets instead of my proposal. Maybe I should add a ticket so we can tag this for 4.x and maybe discuss on a developer meeting?

Issue #6782, opened 2014-03-07
Empty permissions are not inherited in page


aschempp:
I have discovered a permission issue. Steps to reproduce in demo:

1. Go to page tree
2. Edit page courses
3. Remove all checkboxes for the backend user permissions (basically, none of the selected users should be able to change anything)
4. Login as limited user
5. You will see that permissions are correct on the "Courses" page, but not inherited to the child pages

I wanted to remove access to a specific part of the page for all regular users (because there are only news modules)...

leofeyer:
As discussed on Mumble on May 21st, this needs to be fixed; in version 3.5 and 3.2 (transition phase).

leofeyer:
Fixed in cd37b1be82204e08497dd7a1d8127c1623f78c4d.

PR #7176, opened 2014-07-17
Add unique index for username and email to tl_member

ausi:
It is currently possible to register with an already existing username and/or email if you add a 4-byte UTF-8 character to the end of the username/email.

contao-register

Adding an unique index to the username and email columns would fix this issue. But I don't know if changing the columns to allow NULL values has backwards compatibility issues.

### DIFF
--- a/system/modules/core/dca/tl_member.php
+++ b/system/modules/core/dca/tl_member.php
@@ -36,8 +36,8 @@ $GLOBALS['TL_DCA']['tl_member'] = array
 			'keys' => array
 			(
 				'id' => 'primary',
-				'username' => 'index',
-				'email' => 'index',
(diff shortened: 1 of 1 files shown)


aschempp:
As discussed on Mumble today, the unique index would only solve the problem for this specific case. It would still result in a page title being cut off if someone uses a 4-byte character in the field.

leofeyer:
The actual solution would be to use to a 4-byte collation such as `utf8mb4_general_ci`. However, this cannot be set as default as long as we are still supporting MySQL 5.0 (see #6782).

leofeyer:
Contao 4.5 supports `utf8mb4` now.

PR #7197, opened 2014-07-24
Changed the div-tag of the main content to the HTML5 main-tag

marcobiedermann:
- changed the div-tag of the main content to the HTML5 main-tag


leofeyer:
Changed in contao/core-bundle@36619c88ce0ca29819808e3e87019c24bc4b19be.

PR #7225, opened 2014-08-04
Strip html entities from linkTitle

fatcrobat:
Strip html entities from linkTitle `<a title=""></a>` attribute. When using custom inserttags, that return html entities, the markup will become invalid, because html entities will be rendered inside the attribute value `<a title="<i class="fa fa-pencil"></i> Submit">Submit</a>`


leofeyer:
Still have no idea what you want to accomplish.

PR #7275, opened 2014-09-01
added WAI-ARIA tags to header, main, aside and footer

marcobiedermann:


leofeyer:
@contao/developers @frontendschlampe What do you think?

marcobiedermann:
There also some WAI-Aria tags for events, dates, gallery and navigation. If you like to, I can add them to this pull request and change the title of it

Toflar:
I'm not into that topic so I cannot tell if they're relevant neither if they're correct. But the more of these things are added by default (given they're useful and correct) the better so I don't have to deal with it :-)

aschempp:
I know nothing about aria…

frontendschlampe:
I think you want to mention @MacKP He's the man for the tags! ;-)

MacKP:
It's an nice Idea. But I don't know if we can do it like that :( It's like always: Are these role-Types always the right role? Is the header always role=banner? And the footer is always for contentinfo? And so on... I don't think so. It's (as always) part of the 'Design' to define this as I think.
Would be nice if someone can proof me wrong xD

regards

ausi:
The roles banner, main, complementary and contentinfo are already the default roles for [header](http://www.w3.org/TR/html51/semantics.html#the-header-element), [main](http://www.w3.org/TR/html51/semantics.html#the-main-element), [aside](http://www.w3.org/TR/html51/semantics.html#the-aside-element) and [footer](http://www.w3.org/TR/html51/semantics.html#the-footer-element) according to the W3C. The spec also says:

> In the majority of cases setting an ARIA role and/or aria-* attribute that matches the default implicit ARIA semantics is unnecessary and not recommended as these properties are already set by the browser.

IMO it’s a matter of browser support, if the browser supports the HTML5-Elements the roles aren’t necessary, but if it doesn’t support them they would add useful information e.g. for screen readers.

Looking at [can I use](http://caniuse.com/#feat=html5semantic) I would say if we support IE8 we should add all roles, but if we support IE9+ we should only use `role="main"` (and `<main>` as requested in #7197). I would like to go with IE9+.

MacKP:
@ausi but it's only neccesary to set it, when there is an other role than the default. So I think we can go ahead and set in only for main. When you wan't to support IE8 AND Screenreaders you can set it by your own I think.

regards

marcobiedermann:
A page can have multiple `header` and `footer` elements. In my opinion setting the *main-header* and *main-footer* is only possible via `role` attribute

ausi:
IMO the *main-header* and *main-footer* are determined by the structure of the markup. The spec says „The `header` element represents introductory content for its nearest ancestor [sectioning content](http://www.w3.org/TR/html51/dom.html#sectioning-content) or [sectioning root](http://www.w3.org/TR/html51/semantics.html#sectioning-root) element.“. So if `header` and `footer` have no `article`, `aside`, `nav` or `section` as an ancestor element they are assigned to the `body` and are therefore the *main-header* and *main-footer* of the page.

leofeyer:
I'm closing this in favor of #7197, which contains the changes we want to make.

PR #7289, opened 2014-09-04
Update event_list.xhtml

frontendschlampe:
fix missing "block"-class and teaser output


leofeyer:
This has been fixed meanwhile.

PR #7315, opened 2014-09-13
requiring value in sole field "password" in ModulePersonalData

ghost:
The frontend module for editing a member's personal data is revised to request new password on submit if module is configured to display password field only.


leofeyer:
There is now (in Contao 3.5) a password module, which also requires the old password to set a new one.

PR #7379, opened 2014-10-17
Allow to define default flags for inserttags

dmolineus:
It would be useful to force some insert tag flags for own insert tags. I need it to force an insert tag being always refreshed. There is currently no way to do it in the `replaceInsertTags` hook.

This implementation provides an generic implementation where every flag can be forced for an insert tag.

### DIFF
--- a/system/modules/core/config/config.php
+++ b/system/modules/core/config/config.php
@@ -472,3 +472,4 @@ $GLOBALS['TL_ASSETS'] = array
  */
 $GLOBALS['TL_MODELS'] = array();
 $GLOBALS['TL_PERMISSIONS'] = array();
+$GLOBALS['TL_INSERTTAG_FLAGS'] = array();
--- a/system/modules/core/library/Contao/Controller.php
+++ b/system/modules/core/library/Contao/Controller.php
@@ -698,6 +698,12 @@ abstract class Controller extends \System
 			$tag = array_shift($flags);
 			$elements = explode('::', $tag);
 
+            if (isset($GLOBALS['TL_INSERTTAG_FLAGS'][$elements[0]]))
+            {
+                $flags = array_merge($GLOBALS['TL_INSERTTAG_FLAGS'][$elements[0]], $flags);
+                $flags = array_unique($flags);
+            }
+
 			// Load the value from cache
 			if (isset($arrCache[$strTag]) && !in_array('refresh', $flags))
 			{
--- a/system/modules/core/library/Contao/Controller.php
+++ b/system/modules/core/library/Contao/Controller.php
@@ -698,11 +698,11 @@ abstract class Controller extends \System
 			$tag = array_shift($flags);
 			$elements = explode('::', $tag);
 
-            if (isset($GLOBALS['TL_INSERTTAG_FLAGS'][$elements[0]]))
-            {
-                $flags = array_merge($GLOBALS['TL_INSERTTAG_FLAGS'][$elements[0]], $flags);
-                $flags = array_unique($flags);
-            }
+			if (isset($GLOBALS['TL_INSERTTAG_FLAGS'][$elements[0]]))
+			{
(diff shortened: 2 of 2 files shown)


leofeyer:
What's the status of this ticket?

Toflar:
Well, I still think my solution would be the correct one for Contao 4 and would allow what David is trying to do with his PR here :)

leofeyer:
Yes, no doubt about that. But my question was rather "do we really need this" and if so, "who creates the PR"?

Toflar:
I can do a PR when your hooks-to-events-PR is merged because then we already have the event and adding this must be piece of cake, right? ;-)

leofeyer:
I guess so. But I don't really think we need default flags for insert tags at all.

Toflar:
Me neither but @dmolineus's use case for having a `forceRefresh()` method is a valid use case.

leofeyer:
> It would be useful to force some insert tag flags for own insert tags

I just realized that we are talking about **custom** insert tags here. Then we can close the ticket, because the `replaceInsertTags` hook provides all necessary parameters to accomplish it.

```php
$varValue = $this->$callback[0]->$callback[1]($tag, $blnCache, $arrCache[$strTag], $flags, $tags, $arrCache, $_rit, $_cnt);
```

dmolineus:
No, its not. If the insert tag is already called, it's always cached internally and the hook is **not recalled**. 

I looked at the code when opening this issue. Gonna show you the details tomorrow. Right now only airport WiFi.

I'm fine to create the pull request. Guess it's gonna be a part of CTO 4. Which is the related PR which has to be merged before?

leofeyer:
> If the insert tag is already called, it's always cached internally

Not if you prevent caching upon the first call.

Toflar:
> I'm fine to create the pull request. Guess it's gonna be a part of CTO 4. Which is the related PR which has to be merged before?

https://github.com/contao/core-bundle/pull/204

leofeyer:
We don't need a PR. It can be accomplished with the existing hook!

Toflar:
Disagree. I don't want to use 2 hooks when I define a new InsertTag just to accomplish that it is never cached. But you don't have to worry, @dmolineus said he's gonna provide one.

leofeyer:
Why 2 hooks? A custom insert tag will never trigger two hooks (I assume you refer to the "unknown insert tag flag" hook).

I'm going to try it later today and if I get it to work, you owe me a beer ;)

Toflar:
> I'm going to try it later today and if I get it to work, you owe me a beer ;)

I'll have one with you anyway :D

leofeyer:
That's good, because it is eventually not possible to bypass the internal cache using the hook. I figured you could modify the array key, but `$strTag` is not passed as argument.

Still I vote -1 on the suggested feature, because if we add default flags, we'd also have to add a way to override them; and this will make everything more complicated. Apart from this, there is a really simple solution: just add the `|refresh` flag to the insert tag.

Toflar:
I think default flags don't make sense either. The only one that does is `refresh` because right now you cannot define an InsertTag that should never be cached. And that's exactly what we want. I don't want to define default flags, I just want to force refreshing it on every replacement :) So if the event provides a `forceRefresh` option, we're all happy already. No need to talk about the flags at all :)

leofeyer:
> because right now you cannot define an InsertTag that should never be cached

Yes, you can: `{{custom|refresh}}`

Toflar:
Which is exactly what we don't want to have. If `custom` **ALWAYS** needs a refresh, then we don't want the user to use the `refresh` flag everytime. It's pointless if the developer already knows that without the flag his InsertTag is useless :)

leofeyer:
I see the point, but that's exactly why the flag has been added.

dmolineus:
I had a look at my proposed changes again. I agree with @Toflar that default flags is not required. If any flag should applied all the time it can be hardcoded in the custom flag. 

But I would really love seeing an option set the force flag in the InsertTag event as @Toflar proposed. It would improve the handling for editors.

PR #7411, opened 2014-10-31
Check the dependencies when disabling modules

aschempp:
see #6576 

Check each individual commit and make sure to ignore whitespaces, the changes are only marginal.


leofeyer:
Since this is obsolete in Contao 4, I am closing the ticket.

PR #7435, opened 2014-11-12
[3.4-dev] Don't resize uploaded SVG images

ausi:
Resizing SVG images doesn't reduce the file size, so they shouldn't be resized on upload.


leofeyer:
Fixed in 765888fcad6aec038a559a87accff40d2ba87c07.

PR #7465, opened 2014-11-22
Also export/import table tl_files in theme export/import

ausi:
As discussed in #7449

### DIFF
--- a/system/modules/core/classes/Theme.php
+++ b/system/modules/core/classes/Theme.php
@@ -95,6 +95,7 @@ class Theme extends \Backend
 			// Store the field names of the theme tables
 			$arrDbFields = array
 			(
+				'tl_files'           => $this->Database->getFieldNames('tl_files'),
 				'tl_theme'           => $this->Database->getFieldNames('tl_theme'),
 				'tl_style_sheet'     => $this->Database->getFieldNames('tl_style_sheet'),
 				'tl_style'           => $this->Database->getFieldNames('tl_style'),
@@ -412,10 +413,7 @@ class Theme extends \Backend
 			// Load the DCAs of the locked tables (see #7345)
 			foreach (array_keys($arrLocks) as $table)
 			{
-				if ($table != 'tl_files')
-				{
-					$this->loadDataContainer($table);
-				}
+				$this->loadDataContainer($table);
 			}
 
 			$this->Database->lockTables($arrLocks);
@@ -436,7 +434,7 @@ class Theme extends \Backend
 				$table = $tables->item($i)->getAttribute('name');
 
 				// Skip invalid tables
-				if ($table == 'tl_files' || !in_array($table, array_keys($arrLocks)))
+				if (!in_array($table, array_keys($arrLocks)))
 				{
 					continue;
 				}
@@ -548,8 +546,8 @@ class Theme extends \Backend
 							}
 						}
 
-						// Adjust the file paths in style sheets
-						elseif (\Config::get('uploadPath') != 'files' && ($table == 'tl_style_sheet' || $table == 'tl_style') && strpos($value, 'files') !== false)
+						// Adjust the file paths in style sheets and tl_files
+						elseif (\Config::get('uploadPath') != 'files' && ($table == 'tl_style_sheet' || $table == 'tl_style' || ($table == 'tl_files' && $name == 'path')) && strpos($value, 'files') !== false)
 						{
 							$tmp = deserialize($value);
 
@@ -647,8 +645,16 @@ class Theme extends \Backend
 						new \Folder($set['templates']);
 					}
 
-					// Update the datatbase
-					$this->Database->prepare("INSERT INTO ". $table ." %s")->set($set)->execute();
+					if ($table == 'tl_files')
+					{
+						// Update the datatbase
+						$this->Database->prepare("UPDATE ". $table ." %s WHERE path = ?")->set($set)->execute($set['path']);
+					}
+					else
+					{
+						// Update the datatbase
+						$this->Database->prepare("INSERT INTO ". $table ." %s")->set($set)->execute();
+					}
 				}
 			}
 
@@ -723,21 +729,7 @@ class Theme extends \Backend
 		$strTmp = md5(uniqid(mt_rand(), true));
 		$objArchive = new \ZipWriter('system/tmp/'. $strTmp);
 
-		// Add the folders
-		$arrFolders = deserialize($objTheme->folders);
-
-		if (!empty($arrFolders) && is_array($arrFolders))
-		{
-			$objFolders = \FilesModel::findMultipleByUuids($arrFolders);
-
-			if ($objFolders !== null)
-			{
-				foreach ($this->eliminateNestedPaths($objFolders->fetchEach('path')) as $strFolder)
-				{
-					$this->addFolderToArchive($objArchive, $strFolder);
-				}
-			}
-		}
+		$this->addTableTlFiles($xml, $tables, $objTheme, $objArchive);
 
 		// Add the template files
 		$this->addTemplatesToArchive($objArchive, $objTheme->templates);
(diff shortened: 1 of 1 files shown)


system/modules/core/classes/Theme.php
```diff
@@ -548,8 +546,8 @@ protected function extractThemeFiles($arrFiles, $arrDbFields)
 							}
 						}
 
-						// Adjust the file paths in style sheets
-						elseif (\Config::get('uploadPath') != 'files' && ($table == 'tl_style_sheet' || $table == 'tl_style') && strpos($value, 'files') !== false)
+						// Adjust the file paths in style sheets and tl_files
+						elseif (\Config::get('uploadPath') != 'files' && ($table == 'tl_style_sheet' || $table == 'tl_style' || ($table == 'tl_files' && $name == 'path')) && strpos($value, 'files') !== false)
```
leofeyer: Why did you add an additional check for the field name here?

```
($table == 'tl_files' && $name == 'path'))
```

Are there other fields which potentially contain `files` and shall not be changed?

system/modules/core/classes/Theme.php
```diff
@@ -647,8 +645,16 @@ protected function extractThemeFiles($arrFiles, $arrDbFields)
 						new \Folder($set['templates']);
 					}
 
-					// Update the datatbase
-					$this->Database->prepare("INSERT INTO ". $table ." %s")->set($set)->execute();
+					if ($table == 'tl_files')
+					{
+						// Update the database
+						$this->Database->prepare("UPDATE ". $table ." %s WHERE path = ?")->set($set)->execute($set['path']);
```
leofeyer: Why do you update instead of insert?

system/modules/core/classes/Theme.php
```diff
@@ -958,14 +989,14 @@ protected function addTableTlImageSize(\DOMDocument $xml, \DOMElement $tables, \
...
-	protected function addDataRow(\DOMDocument $xml, \DOMElement $table, \Database\Result $objData, array $arrOrder=array())
+	protected function addDataRow(\DOMDocument $xml, \DOMElement $table, array $arrRow, array $arrOrder=array())
 	{
 		$t = $table->getAttribute('name');
 
 		$row = $xml->createElement('row');
 		$row = $table->appendChild($row);
 
-		foreach ($objData->row() as $k=>$v)
+		foreach ($arrRow as $k=>$v)
```
leofeyer: Why is this change required?

system/modules/core/classes/Theme.php
ausi: An extension may add additional columns to `tl_files` which we shoudn't modify IMO.

system/modules/core/classes/Theme.php
ausi: Because the entries were already inserted by the DBAFS

system/modules/core/classes/Theme.php
ausi: To be able to call this method from `addFolderToArchive()` at line 1139

system/modules/core/classes/Theme.php
leofeyer: Really? We shall only fix our own fields and leave third-party developers with the wrong paths?

ausi: IMO we should only change fields we know which data is in it. Just because a column value starts with `files/` doesn't ensure that it's a path that should be changed.

ausi: A third-party developer should use `$GLOBALS['TL_HOOKS']['extractThemeFiles']` to implement such transformations.

leofeyer:
There is a big problem: You cannot use the `FilesModel` if the table is locked. See here:

https://github.com/contao/core/blob/c6ce781564307040655a85fb48834c8e615e41ed/system/modules/core/classes/Theme.php#L578

You have to use the `Database` class instead. And then you won't have to change line 999 either.

ausi:
Aren’t they sharing the same database object? Why are the locked tables a problem?

leofeyer:
I don't remember :( I only stumbled upon my comment, therefore I figure there must be an issue.

ausi:
The comment was added in 8b71567fef283d7d7630538918474176ba71d9b1, but i don’t see a reason for it.

Is this related to #7345?

Maybe we should remove `if ($table != 'tl_files')` in [system/modules/core/classes/Theme.php:385](https://github.com/contao/core/blob/f8f5d3c72e0a4eb84f6367ed08d19d50d9ead483/system/modules/core/classes/Theme.php#L385).

ausi:
I just took a closer look. Where did I use the `FilesModel` class while tables are locked? I only found occurrences of it in the export methods which don’t lock the tables.

leofeyer:
Here's the adjusted version: https://github.com/contao/core/compare/develop...feature/export-file-paths

@ausi Please review and check whether it works for you.

ausi:
I still don’t like converting all columns of `tl_files` with `customizeUploadPath()`. If you have a column in `tl_files` containing a serialized array like `a:1:{s:1:"a";a:1:{s:1:"b";a:1:{s:1:"c";s:13:"foo files bar";}}}` it would be converted to `a:1:{s:1:"a";a:1:{s:1:"b";s:5:"Array";}}` which is bad. E.g. `tl_files.meta` would also be broken if the array was nested one level deeper.

Your implementation is a bit different than mine: If `tl_files` is out of sync, you export nonexistent file entries, old hash values and don’t export file entries for files which are missing in the database. IMO it would be great to be consistent with the files we are exporting, that’s why I wrote it that way.

leofeyer:
> If you have a column in `tl_files` containing a serialized array

We have this problem in the other tables, too. But it is impossible to foresee which columns people add, therefore we either have to fix all columns or agree on exporting only a defined set of fields (e.g. `path`, `meta` and the important part fields).

> is out of sync, you export nonexistent file entries, old hash values and don’t export file entries for files which are missing in the database

We should rather add a warning to sync the file system instead of trying to work around this IMHO.

ausi:
> > If you have a column in `tl_files` containing a serialized array

> We have this problem in the other tables, too. But it is impossible to foresee which columns people add, therefore we either have to fix all columns or agree on exporting only a defined set of fields (e.g. `path`, `meta` and the important part fields).

I don’t get it. If someone uses `inputType`s that need conversion we already convert the values for them in [Theme.php:353](https://github.com/contao/core/blob/336c6a4e2ace2c878a797d23211140588b72b832/system/modules/core/classes/Theme.php#L535), [553](https://github.com/contao/core/blob/336c6a4e2ace2c878a797d23211140588b72b832/system/modules/core/classes/Theme.php#L553), [574](https://github.com/contao/core/blob/336c6a4e2ace2c878a797d23211140588b72b832/system/modules/core/classes/Theme.php#L574), [989](https://github.com/contao/core/blob/336c6a4e2ace2c878a797d23211140588b72b832/system/modules/core/classes/Theme.php#L989) and [1004](https://github.com/contao/core/blob/336c6a4e2ace2c878a797d23211140588b72b832/system/modules/core/classes/Theme.php#L1004). If someone adds a column with no `inputType` we can’t know which data it stores and therefore we shouldn’t modify the data of such columns, but just export it as it is. An extension developer has to convert special data himself via the `extractThemeFiles` and `exportTheme` hooks, as it is already the case for all other tables and value types, e.g. in [rocksolid-custom-elements/src/MadeYourDay/Contao/CustomElements.php:1271-1375](https://github.com/madeyourday/contao-rocksolid-custom-elements/blob/32041284353305cc162cbf3651448d680253a710/src/MadeYourDay/Contao/CustomElements.php#L1271-L1375). Why should we change that now for `tl_files`?

ausi:
@leofeyer should we discuss this issue in the contao call next week?

leofeyer:
Merged in c03b588dcee6ce23f7effa0ff5a5f77863b34a30.

Issue #7400, opened 2014-10-24
FR: Debug Bar actions in plain javascript instead of mootools


tim-bec:
If you use jquery in your project and activate the debug mode, mootools is loaded to open / hide the debug bar. 
This behaviour took me some time, while i tried to debug my project javascript (jquery). So it would be great to use plain javascript instead of mootools to prevent side effects.

https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Template.php#L343

leofeyer:
Since we are using the Symfony web profiler as of Contao 4, I guess we can close this ticket.

PR #7467, opened 2014-11-22
Refactored debug toolbar to use plain javascript #7400

mruettgers:


leofeyer:
As discussed in Mumble on February 26th, we would merge this pull request if you added IE8 support. However, please be aware that we will no longer use the debug bar in Contao 4 in favor of the Symfony debug bar. So if you decide not to invest any more time, that's fine, too.

bezin:
I wanted to open the exact same issue some minutes ago. Having a debug bar not depending on mootools would have been great for 3.5, since I'll work a lot with this version for at least the next 18 months... If there is a chance, that a plain javascript debug bar will be merged into Contao 3.5.x, I would take care of providing IE8+ support, as long as @mruettgers is fine with that?

mruettgers:
Of course I am fine with that.

PR #7493, opened 2014-12-03
Add a post authenticate hook

dmolineus:
Sometimes it is really useful to have a callback when a user is authenticated. It is triggered all times the user is authenticated no matter if he just logged in or is authenicated from the session cookie.

I have to provide a permission service which gets some data from the authenticated user. So it's not enough using the postLogin hook. 

Using the `initializeSystem` hook and call `User::authenticate` by my own is a bad idea, because it would force every script to require a logged in user.

### DIFF
--- a/system/modules/core/library/Contao/User.php
+++ b/system/modules/core/library/Contao/User.php
@@ -202,7 +202,18 @@ abstract class User extends \System
 		$this->Database->prepare("UPDATE tl_session SET tstamp=$time WHERE sessionID=?")
 					   ->execute(session_id());
 
-		$this->setCookie($this->strCookie, $this->strHash, ($time + $GLOBALS['TL_CONFIG']['sessionTimeout']), null, null, false, true);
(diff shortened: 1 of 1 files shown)


dmolineus:
Any decisions about it? Would love to see it in Contao 3.5.

Toflar:
No objections.

aschempp:
I don't think you should change line 205 in your PR, and I'm not sure about using `objAuth`. Otherwise :+1:

dmolineus:
> I don't think you should change line 205 in your PR, and I'm not sure about using objAuth. Otherwise :+1:

I don't know why Git thinks I've changes this line. Actually that's how it's currently implemented.
https://github.com/contao/core/blob/develop/system/modules/core/library/Contao/User.php#L288

leofeyer:
Merged in 3dbf44b9ffba7eda09a13cb5512d33ad1e69c5e6.

PR #7496, opened 2014-12-03
Important part fields: make file extension test case insensitive

janekbettinger:
Show important part fields for images independent of the case sensitivity of the file extension.


leofeyer:
Merged in df214b55d61c3c5c772ccd304bb4cd5515290f8e.

PR #7497, opened 2014-12-03
Make BackendUser::authenticate() return true to honor documentation.

discordier:
Before the method was returning null which is in contrast to the implementation in User and FrontendUser classes.

### DIFF
--- a/system/modules/core/classes/BackendUser.php
+++ b/system/modules/core/classes/BackendUser.php
@@ -179,7 +179,7 @@ class BackendUser extends \User
 		// Do not redirect if authentication is successful
 		if (parent::authenticate() || \Environment::get('script') == 'contao/index.php')
 		{
-			return;
+			return true;
 		}
 
 		$strRedirect = 'contao/';


leofeyer:
This change is actually wrong, because the method would return `true` even if the user could **not** be authenticated (but `TL_SCRIPT == 'contao/index.php' is true). Also, the return value is never used.

aschempp:
Be aware that we already patched this in Contao 4 (see https://github.com/contao/core-bundle/commit/60db487c70a51f3f320e4c2af4f388990fba9aac)

leofeyer:
@aschempp Thanks for the heads up. The patch in Contao 4 is actually correct and should be back ported here.

discordier:
You are right about the url evaluating. Feel free to close and backport from C4.

Zeromax:
isn't this related? #6850

discordier:
Related but with a bigger scope.

leofeyer:
Fixed in dc7116852eee8af9f5cbc6397612578d5c39e74e.

PR #7504, opened 2014-12-08
frontend preview for news articles

wtom:
just a little feature to use the frontend preview on news articles

### DIFF
--- a/system/modules/news/modules/ModuleNews.php
+++ b/system/modules/news/modules/ModuleNews.php
@@ -288,7 +288,7 @@ abstract class ModuleNews extends \Module
 	 * @param boolean
 	 * @return string
 	 */
-	public function generateNewsUrl($objItem, $blnAddArchive=false)
+	public static function generateNewsUrl($objItem, $blnAddArchive=false)
 	{
 		$strCacheKey = 'id_' . $objItem->id;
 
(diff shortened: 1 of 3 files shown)


system/modules/news/modules/ModuleNews.php
```diff
@@ -288,7 +288,7 @@ protected function getMetaFields($objArticle)
 	 * @param boolean
 	 * @return string
 	 */
-	protected function generateNewsUrl($objItem, $blnAddArchive=false)
+	public function generateNewsUrl($objItem, $blnAddArchive=false)
```
leofeyer: This should be `public static function` otherwise it will trigger a "non-static function cannot be called statically" error.

Toflar:
That would be cool for events as well but the code cannot work, as the `generateNewsUrl()` currently also has some object references in it (e.g. `$this->generateFrontendUrl()`). Also, it should be moved to `News` instead of residing on `ModuleNews`.

jenda77:
+1

leofeyer:
Implemented in 43797a4cdd69c58111d4f520614b4017d77dff91, contao/calendar-bundle@5b4f7052a33d992e059a30645dbd5fe5c7e79749 and contao/news-bundle@6e83434e648bff403e6e1effffd823320f1dd9c1.

PR #7514, opened 2014-12-11
ce_content :: added save callback to youtube field to extract video id from a url

uvoelkel:
Quite a few users struggle with extraction of the video id from a youtube url. Especially with the recently added forced playlist.
This patch adds a save callback to the ce_content.youtube field that extracts the video id from a complete url. The user can just copy and paste whatever the address bar says.

### DIFF
--- a/system/modules/core/dca/tl_content.php
+++ b/system/modules/core/dca/tl_content.php
@@ -627,7 +627,7 @@ $GLOBALS['TL_DCA']['tl_content'] = array
 			'label'                   => &$GLOBALS['TL_LANG']['tl_content']['youtube'],
 			'exclude'                 => true,
 			'inputType'               => 'text',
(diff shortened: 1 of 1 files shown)


leofeyer:
Will this PR ever be completed or can I close the ticket?

u-voelkel:
Sorry, I completely forgot about this one :(
Any other format I am missing?

fritzmg:
I usually use this regex: http://stackoverflow.com/questions/2936467/parse-youtube-video-id-using-preg-match/6382259#6382259

leofeyer:
Implemented in contao/core-bundle@afeef802c5cee9982c0de7b80b2f3bc2053c22a2. The YouTube element now also uses its own template and is embedded via iframe by default.

PR #7516, opened 2014-12-11
Files model now returns a single model when searching by ID

qzminski:
This fix prevents the `findByPk()` and `findById()` method from returning the `\Model\Collection` instance.


leofeyer:
Merged in 3a9da9b850137892ecae90a8b2553ec259f1a15e.

PR #7539, opened 2014-12-25
Add missing rawurldecode, fixes missing image width and height

ausi:
If you select an image with a space in the filename the image width and height attributes are empty:

``` html
<img src="files/foo%20bar.jpg" width="" height="" alt="">
```

Similar issue with the file insert tag:

``` html
<img src="files/foo%20bar.jpg"  alt="">
```

This PR fixes both issues.

Related: 758c1d1b9358f73a5b3115d46f82fcd7f43a3bcb

I also added a test in contao/core-bundle#25


leofeyer:
Merged in f92ccfee388c39f49ad9ff2f597233076b097ef8.

PR #7549, opened 2015-01-07
Remove inline styles from image captions

ausi:
The inline styles for image captions don’t make sense anymore. Because of the new responsive images feature the image width is dynamic.

Related: #3517, [Forum thread 54806](https://community.contao.org/de/showthread.php?54806#post353565)

### DIFF
--- a/system/modules/core/templates/elements/ce_accordion.html5
+++ b/system/modules/core/templates/elements/ce_accordion.html5
@@ -25,7 +25,7 @@
         <?php endif; ?>
 
         <?php if ($this->caption): ?>
-          <figcaption class="caption" style="width:<?php echo $this->arrSize[0]; ?>px"><?php echo $this->caption; ?></figcaption>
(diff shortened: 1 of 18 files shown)


leofeyer:
Did we not solve this with #7351? And doesn't your change revert the fixes from #3517?

ausi:
> Did we not solve this with #7351?

No, in #7351 we fixed the inconsistency between `$this->arrSize` / `$this->src` and `$this->picture['img']['width']` / `$this->picture['img']['src']`.

As commented in https://github.com/contao/core/issues/3517#issuecomment-23944075 the inline styles don’t work great for responsive websites. And because of the new responsive images feature the width of the image is dynamic and the caption width may not match the width of the image.

leofeyer:
Actually merged in 35fe8d9ee60a4574f24a2bb4597c1c5eab276380.

PR #7555, opened 2015-01-09
ce_downloads - Add templatevariable for direct href-link to File

morhai:
I wanted to directly link to a file with ce_downloads without missing its features (Fileicon, filesize, etc.).
This could be easily done with an additional template variable which gives the direct Link to the file (files/folder/xxxx.xyz) without xxxx?file=files/folder/xxxx.xyz.
I call this var "hreffile".

Edit your ce_downloads Template like that:

```
<?php $this->extend('block_searchable'); ?>
<?php $this->block('content'); ?>
  <?php echo $this->v2warning; ?>
  <ul>
    <?php foreach ($this->files as $file): ?>
      <li><img src="<?php echo $file['icon']; ?>" width="18" height="18" alt="" class="mime_icon"> <a href="<?php echo $file['hreffile']; ?>" title="<?php echo $file['title']; ?>"><?php echo $file['link']; ?> <span class="size">(<?php echo $file['filesize']; ?>)</span></a></li>
    <?php endforeach; ?>
  </ul>
<?php $this->endblock(); ?>
```

This gives more freedom how to display the downloads.

Already requested/discussed here:
https://community.contao.org/de/showthread.php?54333
https://community.contao.org/de/showthread.php?54376

### DIFF
--- a/system/modules/core/elements/ContentDownload.php
+++ b/system/modules/core/elements/ContentDownload.php
@@ -101,10 +101,12 @@ class ContentDownload extends \ContentElement
 		}
 
 		$strHref .= ((\Config::get('disableAlias') || strpos($strHref, '?') !== false) ? '&amp;' : '?') . 'file=' . \System::urlEncode($objFile->value);
+		$strHrefFile = \System::urlEncode($objFile->value);
 
 		$this->Template->link = $this->linkTitle;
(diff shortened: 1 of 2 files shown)


leofeyer:
The whole point of the download(s) element is **not** to have direct links, so one can control the access to the files. For a direct link, you can uns the hyperlink element or simply a link in TinyMCE.

PR #7567, opened 2015-01-14
Always return the model in File and Folder classes

qzminski:


discordier:
Is this fix intended for #7490?

qzminski:
No, not at all. The problem with this issue is that the ```File``` and ```Folder``` classes have both the ```getModel()``` method which returns the model **only** when the operation on the object has been performed (e.g. folder/file was created). Otherwise it is always null. Consider the examples:

```php
// folder does not exist yet
$folder = new \Folder('files/my_folder');
$folder->getModel(); // returns \FilesModel

// folder already exists
$folder = new \Folder('files/my_folder');
$folder->getModel(); // returns null
```

discordier:
Ah, you are right. I was wondering as someone in the community linked to this PR as a fix for the other issue.
I wonder however if this change is a be break as it now returns a model and one then has to check existence by hand. It would be a pretty awkward way of checking for existence anyway, therefore I tend to merge this. Therefore a +1 from me.

Toflar:
> Therefore a +1 from me.

Same here.

leofeyer:
This is actually a BC break, isn't it? People might have checked for `null === $folder->getModel()`.

discordier:
@leofeyer which means the folder exists. Therefore I said it is an awkward check compared to using the methods in the File and Folder class and IMO bad coding habit. We can safely assume this usage of checking as glitch riding, can't we?

leofeyer:
I guess so. Let's just vote on whether to change it in version 3.2.x, 3.5 or 4.0.

@contao/developers Please vote and/or comment.

ausi:
3.5

aschempp:
3.5 for me too. I think every "bug fix" we do is a potential BC break if people currently work around it. Still I think this case is wrong and should be changed.

Toflar:
3.2

discordier:
3.2 because relying on the previous behaviour is also to be considered a bug. :)

Toflar:
> because relying on the previous behaviour is also to be considered a bug. :)

...and should have been reported as a bug instead of incorporated into the code. That's my point of view as well. :)=

qzminski:
Why was this closed?

discordier:
Beacause of the release of 3.4.1.
Reopen the PR agains hotfix/3.4.2 please.

aschempp:
It should be against master ;-)

qzminski:
I think you can always cherrypick my comment.

leofeyer:
I have created a reminder: #7591

3.5 seems a reasonable decision. We have learned once more today that even the tiniest change can have a huge effect (which is why I had to release version 3.4.2 today, too).

leofeyer:
Merged in 573094f452997d7ab954b8c579d473becbfe56ef.

PR #7568, opened 2015-01-14
Redirect back did not work in login module

Toflar:
The redirect back functionality of the login form cannot work because the referer is always reset to the login page when submitting the login form :)


discordier:
Nice catch. :)

leofeyer:
Really? Because it does work the way it is. Here's how you can try:

1. Open the Contao theme store at https://themes.contao.org/de/themes/insurance-broker.html
2. Add the theme to the cart.
3. Click the "proceed to checkout" button.

You will then be redirected to the login screen and once you have logged in, you are correctly being redirected to the checkout module. The redirect back routine is working in this case.

Toflar:
Well, it does not work for me. Have a look at the code, why would you need to override the session value during the form submit post request? :D=

leofeyer:
Hm, I have tried it with and without your patch and it did not make any difference. Can you point me to a scenario where I actually see the difference?

Toflar:
First of all my scenario:

1. I'm  not logged in, and I surf to a product detail page (e.g. "product.html") that only allows me to add the product to the cart when logged in.
2. I click on "Login" which redirects me to a __separate__ page with the login module ("login.html").
3. I enter my credentials and I get redirected to "login.html" again, so it feels the same as a reload. Instead, I'd expect to be redirected to the product detail page ("product.html").

Things I wonder:

1. Can you explain me the use of `$_SESSION['LAST_PAGE_VISITED']`? Why is there anything stored into that session value? I can't see **any** line of code where this would be used outside of `ModuleLogin::generate()` where it's **always** set for every request anyway :-)

2. Whenever I log in, doesn't `FrontendUser::setUserFromDb()` **always** replace my session with what is stored in the DB and thus kill my referer as well?

leofeyer:
Fist the answers to your questions:

Number 2 is actually the answer to number 1. Since the session data is being replaced upon login, we cannot store the last page visited in `$_SESSION['FE_DATA']`, which is why we are using a separate variable.

Things I (also) wonder:

1. Why is the variable set for every request? I don't understand either. The variable will be overwritten with the login page upon the first request already, even before we get to handling the POST data.

2. But why does it work on themes.contao.org? Did you try it there?

Toflar:
> But why does it work on themes.contao.org? Did you try it there?

Did you activate the cache so `FrontendIndex::outputFromCache()` gets called?

leofeyer:
No, the cache timeout is `0` for the login page.

Toflar:
Not the login page, the page you visit before you go to the login page. So the theme detail in this case.

leofeyer:
No, no cache for the cart page either.

Toflar:
Imho, the referer is **never** set to the session, except for the following 3 use cases:

* You are logged in as a front end user
* You are logged in as a back end user
* The page is cached and `FrontendIndex::outputFromCache()` is called

Very strange, need to discuss this on Mumble, I guess :)

leofeyer:
Actually merged in 5f0ef4908528a10d34e3fd798b97fa2304985221.

PR #7572, opened 2015-01-15
Vastly improved performance by only calling the error handler if needed

Toflar:
This pull request improves the performance of **all** Contao installations by an incredible amount of time! I noticed gains in speed of around `50 ms` for a regular Contao setup and up to `100 ms` for setups with a few extensions!!
### How come?

Contao uses a custom error handler named `__error()` and does set it in the `initialize.php` using `set_error_handler()`. Within that handler, it ignores notices (`E_NOTICE`) completely.
However, whenever a notice is raised, the custom error handler is called, because it's set to handle **all** types of errors:

``` php
set_error_handler('__error');
```

Considering that we're talking about thousands (!!) of notices, mainly caused by `Undefined index` notices because of a missing `isset()`, this also equals thousands of calls of `__error()`!
### Solution

`set_error_handler()` accepts a second parameter where you can specify the error levels you want to register your custom handler for. In our case, we only need it to register for the ones specified in the configuration:

``` php
set_error_handler('__error', Config::get('errorReporting'));
```

Boom! All the calls for notices have gone and with them a few thousand - completely redundant - calls of `__error()`!

_Note:_ I had to move the `set_error_handler()` call down a bit so that `Config::preload()` is called first.
### Credits

I was able to spot that issue thanks to the [PHP profiler blackfire.io](https://blackfire.io) and I would like to thank the developers at SensioLabs for making this excellent product available to everyone!

### DIFF
--- a/system/initialize.php
+++ b/system/initialize.php
@@ -61,10 +61,11 @@ require TL_ROOT . '/system/helper/exception.php';
 
 
 /**
- * Set the error and exception handler
+ * Set the default error and exception handlers to make sure
+ * nothing is printed out to the world
  */
-@set_error_handler('__error');
-@set_exception_handler('__exception');
+set_error_handler('__error');
+set_exception_handler('__exception');
 
 
 /**
@@ -89,6 +90,11 @@ require TL_ROOT . '/system/modules/core/library/Contao/ModuleLoader.php';
 class_alias('Contao\\ModuleLoader', 'ModuleLoader');
 
 Config::preload(); // see #5872
+
+/**
+ * Adjust the error handler to the desired error types
+ */
+set_error_handler('__error', Config::get('errorReporting'));
 
 
 /**


qzminski:
+1

dawehner:
Or in other words, it is a bad idea to ignore notices in the first place

guiwoda:
I saw this PR by a retweet from blackfire.
Very interesting situation here, congrats on figuring it out!

I do agree with @dawehner though, NOTICEs should be taken care of as well, at least in development environments.

One suggestion: be careful on what will happen with your code now that you load your configurations before setting a custom error handler. That could lead to errors in configuration files being printed out to the world!

aschempp:
> be careful on what will happen with your code now that you load your configurations before setting a custom error handler.

The error handler is still set first?

Toflar:
> Or in other words, it is a bad idea to ignore notices in the first place

Of course it is, I guess nobody's going to argue otherwise :) Afaik, I did not mention you should ignore them :)

> One suggestion: be careful on what will happen with your code now that you load your configurations before setting a custom error handler. That could lead to errors in configuration files being printed out to the world!

Thanks for the hint and for caring! I guess you're right. Need to think about that :)

aschempp:
Sorry, got your point. You're referring to the Contao config class and not the error report config ;-)

nicolas-grekas:
Btw, [here is the blackfire.io comparison that show the impact of the patch](https://blackfire.io/profiles/compare/b6e2e3d9-10f7-433c-b2b4-bcc9bd087413/graph?callname=__error&selected=__error&settings%5Bdimension%5D=wt&settings%5BcallerDepth%5D=2&settings%5BcalleeDepth%5D=&settings%5BminNodeCost%5D=5&settings%5BminCallCost%5D=0.01).
(See also https://twitter.com/blackfireio/status/555755934036033536)

leofeyer:
Great find, @Toflar! However, the solution is not yet perfect.

1. Why did you move the exception handler? There are quite a few places where exceptions might be thrown between where the function call was and where it is now (e.g. during `Config::preload()`). These exceptions will now be uncaught, which means readable even if "display errors" is disabled.

2. Would it be an option to set the error handler once at the very beginning of the script and a second time after the configuration has been preloaded? Or will it render the performance gain redundant?

Toflar:
> Why did you move the exception handler? There are quite a few places where exceptions might be thrown between where the function call was and where it is now (e.g. during Config::preload()). These exceptions will now be uncaught, which means readable even if "display errors" is disabled.

Just because they were together already but of course we could leave that where it was :)

> Would it be an option to set the error handler once at the very beginning of the script and a second time after the configuration has been preloaded? Or will it render the performance gain redundant?

I think we could do that :) The performance gain will still be quite big as most notices are not thrown during the preload of the configuration but later on.

Toflar:
@leofeyer like that?

leofeyer:
Yes, like that. And I also like that :)

leofeyer:
Merged in be0cef6dcc91aaf616e46bd10f26b61e30a349d2.

leofeyer:
Actually, it does not work for me. I'm now getting:

```
Notice: Undefined index: requires in contao/core/system/modules/core/library/Contao/ModuleLoader.php on line 153
```

It seems logical to me, because now the custom error handler does not handle notices anymore so the default PHP handler does.

leofeyer:
@contao/developers Please have a look.

Toflar:
Do you override 
```php
error_reporting(Config::get('errorReporting'));
```
again?

leofeyer:
No. IMHO the call needs to be in line 224 after the `error_reporting()` call:

```php
/**
 * Configure the error handling
 */
@ini_set('display_errors', (Config::get('displayErrors') ? 1 : 0));
error_reporting((Config::get('displayErrors') || Config::get('logErrors')) ? Config::get('errorReporting') : 0);


/**
 * Adjust the error handler to the desired error types
 */
set_error_handler('__error', Config::get('errorReporting'));
```

First disable reporting of notices, then allow the error handler to ignore them.

leofeyer:
Would this change still provide the performance gain?

aschempp:
Just make sure it's set before `Config::getInstance()` is called (and all custom config files are loaded)

leofeyer:
We could do this right after `Config::preload()`:

```php
/**
 * Adjust the error handling
 */
@ini_set('display_errors', (Config::get('displayErrors') ? 1 : 0));
error_reporting((Config::get('displayErrors') || Config::get('logErrors')) ? Config::get('errorReporting') : 0);
set_error_handler('__error', Config::get('errorReporting'));
```

However, then the error reporting would no longer be configurable from the module configuration, which is only loaded during `$config->initialize()`. Don't know if this matters though, as you could still call the functions directly.

aschempp:
> However, then the error reporting would no longer be configurable from the module configuration, which is only loaded during $config->initialize(). Don't know if this matters though, as you could still call the functions directly.

Thought about this too, but I don't think it makes sense to configure this in the module anyways.

leofeyer:
Unless someone wants to create a "configure your error reporting behavior in the back end settings" module :)

aschempp:
Not really, that would use the File driver and store to localconfig.

leofeyer:
Changed in db8fd282889e536769ee6f02eda67abd0d9fbe8a then.

PR #7576, opened 2015-01-16
add .gitignore-files to all subdirectories of assets/images to make sure they are present in git.

dtomasi:
If you download a package form contao.org, initialize a new git-repo for the current project and push it up to a git-repo-server, the folders will not be present anymore. So if you deploy a project based on that remote-repository you will run in problems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ nbproject/
 /contao/update.phar.php
 /files/*
 /assets/css/*
-/assets/images/*
 /assets/js/*
 /share/*.xml
 /system/cache/*
@@ -27,9 +26,7 @@ nbproject/
 /system/config/localconfig.php
 /system/config/pathconfig.php
 /system/cron/cron.txt
-/system/logs/*
 /system/modules/!composer
 /system/modules/*/.skip
-/system/tmp/*
 /templates/*
 /vendor/
--- a/assets/images/0/.gitignore
+++ b/assets/images/0/.gitignore
@@ -1,0 +1,3 @@
+# Dummy file to make Git create the folder
+.gitignore
+!.gitignore
--- a/assets/images/0/index.html
+++ b/assets/images/0/index.html
@@ -1,9 +1,0 @@
-<!DOCTYPE html>
-<html lang="en">
-<head>
-<meta charset="utf-8">
-<title>Blank page</title>
-</head>
-<body>
-</body>
-</html>
--- a/assets/images/1/.gitignore
+++ b/assets/images/1/.gitignore
@@ -1,0 +1,3 @@
+# Dummy file to make Git create the folder
+.gitignore
+!.gitignore
--- a/assets/images/1/index.html
+++ b/assets/images/1/index.html
@@ -1,9 +1,0 @@
-<!DOCTYPE html>
-<html lang="en">
-<head>
-<meta charset="utf-8">
-<title>Blank page</title>
-</head>
-<body>
-</body>
-</html>
--- a/assets/images/2/.gitignore
+++ b/assets/images/2/.gitignore
@@ -1,0 +1,3 @@
+# Dummy file to make Git create the folder
+.gitignore
+!.gitignore
--- a/assets/images/2/index.html
+++ b/assets/images/2/index.html
@@ -1,9 +1,0 @@
-<!DOCTYPE html>
-<html lang="en">
-<head>
-<meta charset="utf-8">
-<title>Blank page</title>
-</head>
-<body>
-</body>
-</html>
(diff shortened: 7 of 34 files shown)


leofeyer:
Why are you adding `.gitignore` files in these folders? There is an `index.html` file already, which should suffice so the folder is being created.

dtomasi:
the index.html is ignored by the .gitignore on project-root-level (/assets/images/*). So if you clone this repo, reset it and begin a project on this base, the subdirectories will be ignored. In our project we don´t want to have the whole /assets/images folder, but all its empty subdirectories.

dtomasi:
I think should talk a bit more about the background.
We have a own Git-Repository Server which includes a Continuous Integration for Contao-Projects. For the project setup we have a generator-script that clones this repo, reset it to the needed release-tag, includes the german language-flies via `tx pull` and prepare the installation for our servers. If we do this, I have to add these `.gitignore`´s to all the subdirectories because the `index.html` files would be ignored at root-level `.gitignore`.
I hope this helps a bit more to understand.

Best regards Dominik

dtomasi:
The same happens if you download a release via contao.org, init a empty git for the project and push it to a project remote-repo. The folders will not be present there. I was sometimes running in troubles because of that. So this Request would not only allow us to handle project setup in our way, I think this can help many developers which working with git in their projects.

Best regards Dominik

aschempp:
So the fix should be not to exclude these files in the root gitignore?

dtomasi:
Yes, with this solution we do both. Exclude all image-files which are generated by contao from repo and make sure the folders are present in the repo, so contao can generate them into the folders.

aschempp:
Can you try adding `!/assets/images/*/index.html` into the root `.gitignore` file?

dtomasi:
I tried that, but this does not work for me. 
Anyway, I realized that there is much more to customize for my needs. If you don´t think this is a better solution to make sure the directories are present, you can close this request. I´m going to maintain my adjustments in a separated repo. 

Thanks and best regards
Dominik

aschempp:
Imho we should fix the incorrectly ignored .html files, but your approach is wrong. The change should be in the root .gitignore. If my version does not work, we might have to specify each file individually.

leofeyer:
> Imho we should fix the incorrectly ignored .html files

Even if you ignore a folder, you can force files in it to be part of the git repo. I have done just that with the `index.html` files.

aschempp:
I know, but I feel like this is bad practice. If you download the ZIP file from GitHub and create an installation under version control, the files would be hidden to you.

A better solution would be to create the necessary folders on the fly so we don't need the index.php at all.

discordier:
Afair the index.html is intended to not have an accidentally exposed directory listing when indexes is on in apache. Creation of the folder is only a nice additional effect.

leofeyer:
Exactly.

aschempp:
> Afair the index.html is intended to not have an accidentally exposed directory listing when indexes is on in apache. Creation of the folder is only a nice additional effect.

I thought so too, even though I think it's not really necessary (because the pics are public anyways). Nevertheless, the index.php could be created by the maintenance module same as for the old `html` folder.

Personally, I'd love to see both fixed: update the .gitignore and generate the folders on the fly.

leofeyer:
@contao/developers What needs to be done here?

ausi:
The *.gitignore* file should be changed from

```
/assets/images/*
```

to

```
/assets/images/*/*
!/assets/images/*/index.html
```

And in `Contao\Image::executeResize()` the folder should be created if `dirname(TL_ROOT . '/' . $this->getCacheName())` doesn’t exist.

leofeyer:
> ```
/assets/images/*/*
!/assets/images/*/index.html
```

@ausi Apparently this does not work (see https://github.com/contao/core/pull/7576#issuecomment-70403110).

> And in `Contao\Image::executeResize()` the folder should be created if `dirname(TL_ROOT . '/' . $this->getCacheName())` doesn’t exist.

If we unignore the files, this should not be necessary anymore in Contao 3. And in Contao 4, the folders are created by the install command anyway.

ausi:
>> ```
>> /assets/images/*/*
>> !/assets/images/*/index.html
>> ```

> @ausi Apparently this does not work (see https://github.com/contao/core/pull/7576#issuecomment-70403110).

I tested it and it did work on my machine.

leofeyer:
Fixed in 89e2ac8564fb108e761f6c94726094834fc666b4 then.

ausi:
You should also change `/assets/images/*` to `/assets/images/*/*` to make it work as described in https://github.com/contao/core/pull/7576#issuecomment-155471308

leofeyer:
Fixed.

PR #7586, opened 2015-01-22
Add possibility to attach an original clone to the registry again

Toflar:
I'm a bit on that performance trip lately :)

So I found that on my page there are several identical queries to the database, that look like this:

```
SELECT tl_page.* FROM tl_page WHERE tl_page.id='43' LIMIT 0,1
```

I then found that this is because when you call `PageModel::loadDetails()` or `PageModel::findWithDetails()` (that internally calls `loadDetails()` again), it calls `preventSaving` (see issue https://github.com/contao/core/issues/7199) the model is being detached from the model registry and thus fetched from the database again every time somebody calls `PageModel::findByPk(43)`.

It get's especially costly if somebody calls `PageModel::findWithDetails()` because this will cause Contao do run this very expensive method over and over again.

Needs some good testing and review by @contao/developers, please!

### DIFF
--- a/system/modules/core/library/Contao/Model.php
+++ b/system/modules/core/library/Contao/Model.php
@@ -181,6 +181,21 @@ abstract class Model
 
 
 	/**
+	 * Clone a model with exactly the same original values
+	 *
+	 * @return static
+	 */
+	public function cloneOriginal()
+	{
+		$clone = clone $this;
+
+		$clone->setRow($this->originalRow());
+
+		return $clone;
+	}
+
+
+	/**
 	 * Set an object property
 	 *
 	 * @param string $strKey   The property name
@@ -261,6 +276,29 @@ abstract class Model
 	public function row()
 	{
 		return $this->arrData;
+	}
+
+
+	/**
+	 * Return the original values as associative array
+	 *
+	 * @return array The data record
+	 */
+	public function originalRow()
+	{
+		$row = $this->row();
+
+		if (!$this->isModified()) {
+			return $row;
+		}
+
+		$originalRow = array();
+
+		foreach ($row as $k => $v) {
+			$originalRow[$k] = isset($this->arrModified[$k]) ? $this->arrModified[$k] : $v;
+		}
+
+		return $originalRow;
 	}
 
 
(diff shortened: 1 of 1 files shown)


system/modules/core/models/PageModel.php
```diff
@@ -663,4 +681,70 @@ public function getFrontendUrl($strParams=null, $strForceLang=null)
...
+	{
+		return $this->blnDetailsLoaded;
+	}
+
+	/**
+	 * Set if details are loaded
+	 *
+	 * @param boolean $blnDetailsLoaded
+	 */
+	public function setDetailsLoaded($blnDetailsLoaded)
```
discordier: Should be at least protected IMO.

discordier:
Aside from the line notes.
I somewhat dislike your approach.

I totally agree that the queries should be saved but I'd rather lever the workload to the`detach()` method.

I'd rework it to look like:
```php
	/**
	 * Detach the model from the registry
	 */
	public function detach($blnKeepClone = true)
	{
		\Model\Registry::getInstance()->unregister($this);

		if ($blnKeepClone) {
			$this->cloneOriginal()->attach();
		}
	}
```

IMO this would impose less problematic changes and it should prevent any duplicate query then.
What do you think?

Toflar:
Updated. Regarding this:

> It get's especially costly if somebody calls PageModel::findWithDetails() because this will cause Contao do run this very expensive method over and over again.

We need to find another solution. We cannot pass on the details as that would result in another model that must be prevented from being stored. I guess we'd need something like a `PageDetailsLoader` or so that internally stores the details it has loaded for a given page id?

leofeyer:
> I'm a bit on that performance trip lately :)

Which is good for us.

> I guess we'd need something like a `PageDetailsLoader` or so that internally stores the details it has loaded for a given page id?

We should definitely discuss this next Thursday.

leofeyer:
As discussed on Mumble on January 29th, we should stop overriding the model properties and implement getters for the additional parameters added by `loadDetails()` instead.

```php
$objPage = PageModel::findByPk(14);
$strLang = $objPage->getLanguage(); // <-- triggers the "find details" routine interally
```

discordier:
Regarding an URI builder, may I hint you to have a look at: https://github.com/contao-community-alliance/url-builder/blob/master/src/ContaoCommunityAlliance/UrlBuilder/UrlBuilder.php

leofeyer:
We don't need a classic URI builder (well, we do as part of the solution) but rather a class which finds the root domain and language and internally caches this data to reduce the number of `loadDetails()` calls.

Toflar:
A very simple but ugly solution to the performance problem would be something like this:

```php
// PageModel.php
public function loadDetails()
{
    \PageDetailsLoader::getInstance()->load($this);
}
```

```php
class PageDetailsLoader
{
    private $details = array();

    public function load(\PageModel $model)
    {
        if (isset($this->details[$model->id])) {
            $this->addDetailsToModel($model, $this->details[$model->id]);
            return;
        }

        // Load
    }
}
```

Just as an idea...

discordier:
Why not making this a private property of the model class itself then? At least until we have a proper registry due to doctrine. :)

I know, both are pretty ugly.

aschempp:
Funny, I just noticed there already is a caching in (deprecared) `\Controller::getPageDetails` (see https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Controller.php#L2909-L2947)

leofeyer:
Merged in bf4745dc9354749488f070df7c73390351838447.

Issue #7500, opened 2014-12-05
[3.4.0] Width and height attributes shouldn't be set if the sizes attribute is specified


ausi:
We should change the if-condition for the width and height attributes from `!$this->sources` to `!$this->sources && empty($this->img['sizes'])`.

See also <https://community.contao.org/de/showthread.php?54305>

aFarkas:
Removing the ``width``/``height`` attributes is the right thing to do. 

This is maybe a little bit off topic, but you could add a new feature for developers: [**add instrinsic ratio HTML/CSS**](http://alistapart.com/article/creating-intrinsic-ratios-for-video/)
The reason why self-proclaimed SEO experts are demanding to specify width and height attributes, is because they are running page speed insights on a webpage and using these results to consult with.

In case of adaptive images this page speed insight test doesn't make any sense, but the ratio behind this test can accomplished (minimize reflows by reserving/occupying the width and height space for the image) with the so called CSS intrinsic ratio padding-bottom pattern. Unfortunately this means you need to know the exact image dimension ratio, which isn't often known by the frontend. Therefore it would be huge, if developers could opt-in to let contao generate this markup automatically:

```html
<style>
.contao-intrinsic-box {
    display: block;
    position: relative;
    height: 0;
    overflow: hidden;
}
.contao-intrinsic-box img,
.contao-intrinsic-box video,
.contao-intrinsic-box iframe,
.contao-intrinsic-box object {
    display: block;
    position: absolute;
    top: 0;
    left: 0;
    width: 100%;
    height: 100%;
}
</style>


<span class="contao-intrinsic-box contao-landscape-box" style="padding-bottom: 42.86%">
    <img
        src="http://placehold.it/175x75"
        sizes="foobar"
        srcset="http://placehold.it/175x75 175w,
        http://placehold.it/350x150 350w,
        http://placehold.it/700x300 700w,
        http://placehold.it/1400x600 1400w" />
</span>
```

This would be really huge. There are also a lot of scripts out there (mostly carousels/sliders, but also isotope/masonry) trying to wait until all images are loaded to get the right dimension of their items. These scripts currently epically fail as soon as it comes to responsive images. With the intrinsic ratio pattern those problems are also automatically solved. What do you think? Should I open a new issue for this feature?

Additionally you can also think of adding the height descriptor. While this descriptor isn't in use by any browser yet, there is some discussion to use this to also solve this problem regardless of the full height descriptor implementation:

```html
<img
[...]

PR #7594, opened 2015-01-23
Allow flexible attributes for the image in the picture element.

dmolineus:
At the moment the image of the picture element is limited to a fixed set of classes. It would be nice if all items of the `img` key are rendered as attributes. So it would be possible that f.e. an class is passed from the outside (wrapper template or hook).

This pr provides the implementation for it.

According to the [allowed attributes](http://de.selfhtml.org/html/referenz/attribute.htm#img) there is only the `ismap` attribute which can be empty. So I decided not to check it. If someone really needs it changing of the template is always possible.

### DIFF
--- a/system/modules/core/templates/picture/picture_default.html5
+++ b/system/modules/core/templates/picture/picture_default.html5
@@ -8,7 +8,7 @@
     <!--[if IE 9]></video><![endif]-->
 <?php endif; ?>
 
-<img src="<?php echo $this->img['src']; ?>"<?php if ($this->img['srcset'] !== $this->img['src']): ?> srcset="<?php echo $this->img['srcset']; ?>"<?php endif; ?><?php if (!empty($this->img['sizes'])): ?> sizes="<?php echo $this->img['sizes']; ?>"<?php elseif (!$this->sources): ?> width="<?php echo $this->img['width']; ?>" height="<?php echo $this->img['height']; ?>"<?php endif; ?> alt="<?php echo $this->alt; ?>"<?php if ($this->title): ?> title="<?php echo $this->title; ?>"<?php endif; ?>>
+<img<?php array_walk($this->img, function($value, $key) { echo sprintf(' %s="%s"', $key, $value); }); ?>>
 
 <?php if ($this->sources): ?>
   </picture>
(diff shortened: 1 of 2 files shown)


aschempp:
Your PR would remove all the conditions and add every array property even if the value is empty?

dmolineus:
It's true, if you have a look at the implementation of `Picture::getTemplateDataSource` it would only affect the `srcset` attribute if no image can be displayed.

https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Picture.php#L218

The empty checking which is used before is actually an `isset` checking which is not required when walking over the array.

ausi:
There is also a check around the width and height attributes, see #7500.

> The empty checking which is used before is actually an `isset` checking which is not required when walking over the array.

`Picture::getTemplateData()` may not be the only source for the data passed to the picture template, so the `empty()` check is IMO better.

What about adding an `attributes` variable?

```html+php
<img src="<?php echo $this->img['src']; ?>" … <?php echo $this->attributes; ?>>
```

And use it like

```php
$this->insert('picture_default', array('attributes' => 'usemap="foo" ismap'));
```

dmolineus:
> Picture::getTemplateData() may not be the only source for the data passed to the picture template, so the empty() check is IMO better.

You are right, a valid argument.

> What about adding an attributes variable?

The way your example implement it, it only would be possible to pass it from the parent template. An hook could not add the data. So when just using an additional `attributes` variable, I prefer adding it as additional key to the `img` variable.

```php
<img src="<?php echo $this->img['src']; ?>" … <?php echo $this->img['attributes']; ?>>
```

ausi:
> The way your example implement it, it only would be possible to pass it from the parent template. An hook could not add the data.

Which hook are you referring to?

dmolineus:
> Which hook are you referring to?

I'm thinking of the `parseTemplate` hook. For example when a content element is rendered I want to add some specific data to the image element. So the hook would check `ce_*` templates which have a picture property. I can access the `picture` attribute and so manipulate the `picture['img']` settings.

I cannot hook on the the `picture_*` template because I do not have the template properties I need from the CE.

ausi:
Images are inserted via `<?php $this->insert('picture_default', $this->picture); ?>`. In the `parseTemplate` hook for `ce_*` templates you can modify `$template->picture['attributes']`.

dmolineus:
As I said, if the attributes are set as part of the picture attribute, you are absolutely right and we are talking about the same. 

Just you're example above (`$this->insert('picture_default', array('attributes' => 'usemap="foo" ismap'));`) did not make it clear

ausi:
@leofeyer I think we can close this because #7718 was merged.

PR #7599, opened 2015-01-27
Implemented diff view for templates

Toflar:
This is very helpful to easily diff templates against the original one in the back end. One can see where the differences are with one click only. I often diff when I update Contao installations to see if the changes are still necessary after the update or to know if there is even a difference or I can just delete the template.

This small feature makes it really easy to do so right from within the back end. See the screenshots:

screen shot 2015-01-27 at 18 30 26
screen shot 2015-01-27 at 18 30 38

### DIFF
--- a/system/modules/core/config/config.php
+++ b/system/modules/core/config/config.php
@@ -47,7 +47,8 @@ $GLOBALS['BE_MOD'] = array
 		'tpl_editor' => array
 		(
 			'tables'      => array('tl_templates'),
-			'new_tpl'     => array('tl_templates', 'addNewTemplate')
+			'new_tpl'     => array('tl_templates', 'addNewTemplate'),
+			'diff'        => array('tl_templates', 'diffTemplate'),
 		)
 	),
 
--- a/system/modules/core/dca/tl_templates.php
+++ b/system/modules/core/dca/tl_templates.php
@@ -103,6 +103,21 @@ $GLOBALS['TL_DCA']['tl_templates'] = array
 				'href'                => 'act=source',
 				'icon'                => 'editor.gif',
 				'button_callback'     => array('tl_templates', 'editSource')
+			),
+			'diff' => array
+			(
+				'label'               => &$GLOBALS['TL_LANG']['tl_templates']['diff'],
+				'href'                => 'key=diff',
+				'icon'                => 'diff.gif',
+				'button_callback'     => function($row, $href, $label, $title, $icon, $attributes) {
+					return sprintf('<a href="%s" title="%s" onclick="Backend.openModalIframe({\'width\':768,\'title\':\'%s\',\'url\':this.href});return false"%s>%s</a>',
+						$this->addToUrl($href . '&amp;popup=1&amp;id=' . $row['id']),
+						specialchars($title),
+						specialchars(str_replace("'", "\\'", sprintf($GLOBALS['TL_LANG']['tl_templates']['diff'][1], $row['id']))),
+						$attributes,
+						\Image::getHtml($icon, $label)
+					);
+				}
 			)
 		)
 	),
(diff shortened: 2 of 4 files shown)


aschempp:
++1

fjacobi:
+1

CodingRockstar:
+1 :)

discordier:
I guess to see a bug in here... what happens if the original is not present?
Like: create a copy of `fe_page` and rename it to `fe_page_me`.

I guess it will fail miserably with file not found on trying to retrieve the contents of said file.
Have not tested, so it is a wild guess (I still miss the unit tests :().

frontendschlampe:
good idea @Toflar ... also good idea from @discordier :-)

Toflar:
True, fixed :)

Toflar:
However, one could still compare `fe_page_stupid` against `fe_page`...
Meh, I guess we need to be able to select to which one you want to diff against. The original one can just be the default.

frontendschlampe:
> Meh, I guess we need to be able to select to which one you want to diff against. The original one can just be the default.

That would be great!

Toflar:
Yeah, I'll implement it later this week :)

lionel-m:
:+1:

MacKP:
+1

discordier:
As long as you know what the original one is. We have more than one for certain candidates. Do you want to provide the possibility to compare any against any?

aschempp:
> Do you want to provide the possibility to compare any against any?

I think we could limit at least to the same prefix…

Toflar:
Thought about that too, but that doesn’t work.
Example: `iso_reader_default`  —> `iso_reader_custom`.
They don’t have the same prefix and if you are about to telling me that `iso_reader_` is the prefix, I’ll ask you how to know if `iso_` or `iso_reader_` or `whatever_you_want_to_put_here_to_compare` is the prefix? :) It doesn’t work.=

aschempp:
The prefix is `iso_` and it IS the same. Obviously `iso_reader_` would be "closer" but I know that won't work.

Toflar:
Okay, so the first occurrence of `_` defines the `prefix`.

aschempp:
To me, yes. At least I can't see a better option.

frontendschlampe:
I wouldn't do so. I think it should be more flexible without any prefix. We don't know if the prefixes will exist in the future and I think there's no need to add such a prefix to my templates for an extension!

Toflar:
So it's implemented.

### Use case 1
When your template does not override one from any module (extension or core) and is completely standalone. In this case it's a template named `foobar`. Note the info message that informs the user about this situation. You can then select and compare it to whatever template you like (@frontendschlampe 's use case):

screen shot 2015-01-28 at 12 08 45

### Use case 2
When your template does not override one from any module (extension or core) just because it's name does not match. In this case I am comparing `fe_page_custom` to `fe_page` and this would then look like this:

screen shot 2015-01-28 at 12 14 29

### Use case 3
When your template overrides one from any module (extension or core). The info message tells you again what happens and by default the original template is preselected and immediately compared to (in this case I did not change anything so they're identical). You see that I want to compare my `ce_teaser` template so only `ce_` templates are being shown because of the prefix support. Obviously I can still compare it to `ce_text` which does not make sense at all but I can't fix stupidity :)

screen shot 2015-01-28 at 12 17 25

lionel-m:
If we use the template inheritance feature, we see that all the contents of the template `fe_base.html5` is replaced with for example:

`fe_custom.html5`:
```PHP
<?php $this->extend('fe_base'); ?>

<?php $this->block('content'); ?>
  <p>New content.</p>
<?php $this->endblock(); ?>
```
Or, We can see the result? Could this be possible?

Samuelll:
+1

xantippe:
++1

Toflar:
@lionel-m I don't understand your question.

netzarbeiter:
+1

lionel-m:
For example I will add a new meta tag in the `fe_page` template. I create the `fe_custom` template and I add a new line `<meta ... >`. The comparison function shows me that I have added a new line in green.

But now, I will use the new template inheritance feature. I remove all the content of the `fe_custom` template and I add:

```PHP
<?php $this->extend('fe_base'); ?>

<?php $this->block('head'); ?>
  <?php $this->parent(); ?>
  <meta ... >
<?php $this->endblock(); ?>
```
I did exactly the same thing. I only added a new meta tag. But in this case, normally, the comparison function shows me that between  `fe_page` and  `fe_custom` all the content has been removed and my code above has been added.

But it does not show the change after the conversion, that is, just a new line in green with the meta tag. I thought about that but it's too complicated...

discordier:
How should it be possible to detect that the output will only have the meta added?
You want to see the difference of generated output in as difference of templates.
Even if we implement it like that (by i.e. executing all the extend() calls etc.) you will not be able to detect which changes are reflected by the very template you are diffing currently as the ones you extend might extend another one etc.
In the end you will end up with a lot of lines marked as different but only one is really in the current template and others are in entirely different templates in the hierarchy.

To put long story short, no this is not possible within reasonable effort and will provide almost no benefit at all.

leofeyer:
Revised and merged in 5e896063f58c3e611c24d0ab86416c1d9fadc076.

Toflar:
Thank you, I hope a lot of people will like that addition :+1: :)

leofeyer:
I do :)

Aybee:
Im Ordner **/system/themes/default/images** fehlt noch die Datei **diffTemplate.gif**. Deshalb ist die Option mit dem default Theme noch nicht sichbar.

Ein direkter Vergleich zeigt, dass die Grafiken **arrows.png** und **diffTemplate.gif** aus dem flexible Theme in dem default Theme nicht enthalten sind.

leofeyer:
Hinzugefügt in d163689662669d3f44bc29ffb2a5f253d6d178fb.

PR #7612, opened 2015-02-02
Allow IP ranges in trusted proxies

aschempp:
The `$GLOBALS['TL_CONFIG']['proxyServerIps']` allows for a list of trusted proxy IPs. In my case, I had to whiteliste several subnetworks of proxy servers, so the simple list was only partially useful.

This PR adds support for IPv4 subnetworks in the list of trusted proxies.
Example: `192.168.1.0/24` would white-list IPs ranging from `192.168.1.1` to `192.168.1.254`.

### DIFF
--- a/system/modules/core/library/Contao/Environment.php
+++ b/system/modules/core/library/Contao/Environment.php
@@ -371,15 +371,35 @@ class Environment
 			$arrIps = array($strXip);
 		}
 
-		$arrIps = array_reverse($arrIps);
+		// Filter away the trusted IPs or IP ranges
+		$arrIps = array_filter($arrIps, function($ip) use ($arrTrusted)
+		{
+			foreach ($arrTrusted as $trusted)
+			{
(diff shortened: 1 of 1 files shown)


qzminski:
+1

Toflar:
We could use a library for that. E.g. https://github.com/Jalle19/php-whitelist-check (100% code coverage) :)

discordier:
We should also allow ipv6

leofeyer:
Where is your code from? Did you use the Symfony routines?

http://symfony.com/doc/current/reference/configuration/framework.html#trusted-proxies

aschempp:
No, I used some code from StackOverflow. I agree we could use the component but then we should only add it in Contao 4. That would also mean we would be compatible with the Symfony container definition.

leofeyer:
The setting has been removed in Contao 4, because Symfony now takes care of it.

PR #7616, opened 2015-02-03
Add HOOK manipulateMailMessage to provide base for feature #7310.

discordier:
Now it is possible to manipulate the SwiftMailer message just before it
get's sent.
This is useful when trying to sign messages etc.

Example code for issue #7310:

config.php:

``` php
$GLOBALS['TL_HOOKS']['manipulateMailMessage'][] =
array('ManipulateMessageHandler', 'signMail');
```

ManipulateMessageHandler.php:

``` php
class ManipulateMessageHandler
{
    public function signMail($message)
    {
        $smimeSigner = Swift_Signers_SMimeSigner::newInstance();
        $smimeSigner->setSignCertificate('/path/to/certificate.pem',
'/path/to/private-key.pem');
        $message->attachSigner($smimeSigner);
    }
}
```

### DIFF
--- a/system/modules/core/library/Contao/Email.php
+++ b/system/modules/core/library/Contao/Email.php
@@ -509,6 +509,16 @@ class Email
 		// Set the return path (see #5004)
 		$this->objMessage->setReturnPath($this->strSender);
 
+		// HOOK: manipulate the message
+		if (isset($GLOBALS['TL_HOOKS']['manipulateMailMessage']) && is_array($GLOBALS['TL_HOOKS']['manipulateMailMessage']))
+		{
+			foreach ($GLOBALS['TL_HOOKS']['manipulateMailMessage'] as $callback)
+			{
+				$objHook = (in_array('getInstance', get_class_methods($callback[0]))) ? call_user_func(array($callback[0], 'getInstance')) : new $callback[0]();
+				$objHook->$callback[1]($this->objMessage, $this);
+			}
+		}
+
 		// Send e-mail
 		$intSent = self::$objMailer->send($this->objMessage, $this->arrFailures);
 


system/modules/core/library/Contao/Email.php
```diff
@@ -509,6 +509,16 @@ public function sendTo()
 		// Set the return path (see #5004)
 		$this->objMessage->setReturnPath($this->strSender);
 
+		// HOOK: manipulate the message
+		if (isset($GLOBALS['TL_HOOKS']['manipulateMailMessage']) && is_array($GLOBALS['TL_HOOKS']['manipulateMailMessage']))
+		{
+			foreach ($GLOBALS['TL_HOOKS']['manipulateMailMessage'] as $callback)
+			{
+				$objHook = (in_array('getInstance', get_class_methods($callback[0]))) ? call_user_func(array($callback[0], 'getInstance')) : new $callback[0]();
```
Toflar: You could use `\System::importStatic()`, couldn't you? :)

leofeyer: I stumbled across this as well.

discordier: I could but I see several downsides doing so.

1. There no need to keep the instance after evaluating the hook as static object (stored in system).
2. The static object will be always overridden when I use a custom key (second parameter in `\System::importStatic()` and therefore render the benefit non existant.
3. Not using a custom key will pollute the static array with a whole lot of one time objects.

I can change this if you really want to but simply see no benefit at all.

discordier: @leofeyer Any input on my remarks above?

leofeyer:
As discussed on Mumble on October 14th, we do not want to add new hooks. Please create an event instead. Also, we have asked us if the SwiftMailer service has hooks to configure #7310?

PR #7634, opened 2015-02-11
Improve the IDE compatibility

leofeyer:
I have completed the work once started by @tristanlins to make Contao more compatible with PHP IDEs (mainly getting rid of the "variable accessed through magic property" warnings). The old PR is available here: https://github.com/contao/core-bundle/pull/1

@contao/developers Please check whether you have any objections.


system/modules/comments/classes/Comments.php
```diff
@@ -269,6 +276,11 @@ protected function renderCommentForm(\FrontendTemplate $objTemplate, \stdClass $
 			}
 
 			$arrField['eval']['required'] = $arrField['eval']['mandatory'];
+
+			/**
+			 * @var \Widget $strClass
+			 * @var \Widget $objWidget
```
aschempp: Is this necessary? Does PhpStorm not notice that `new Widget` is of type widget?

system/modules/comments/classes/Comments.php
```diff
@@ -603,7 +618,7 @@ public static function notifyCommentsSubscribers(\CommentsModel $objComment)
 
 		while ($objNotify->next())
 		{
-			// Don't notify the commentor about his own comment
+			/** @var \CommentsNotifyModel $objNotify */
```
aschempp: This should not be necessary if `\CommentsNotifyModel::findActiveBySourceAndParent` has a correct docblock?

system/modules/core/classes/Ajax.php
```diff
@@ -195,6 +196,7 @@ public function executePostActions(\DataContainer $dc)
 			exit;
 		}
 
+		/** @var \DataContainer|object $dc */
```
aschempp: As far as I know, `DataContainer` does not accept arbitrary properties, so this should not be `object` but have `@property` for the available options.

system/modules/core/classes/Backend.php
```diff
@@ -841,7 +848,7 @@ public static function addPageIcon($row, $label, DataContainer $dc=null, $imageA
 		// Add the breadcrumb link
 		$label = '<a href="' . \Controller::addToUrl('node='.$row['id']) . '" title="'.specialchars($GLOBALS['TL_LANG']['MSC']['selectNode']).'">' . $label . '</a>';
 
-		// Return the image
+		/** @var \DataContainer|object $dc */
```
aschempp: This should be in the docblock of the function and not inline code.

system/modules/core/classes/DataContainer.php
```diff
@@ -81,7 +92,7 @@ class DataContainer extends \Backend
 
 	/**
 	 * Active record
-	 * @param object
+	 * @param \Model
```
aschempp: The class properties should be `@var` or `@type` and not `@param` afaik

system/modules/core/classes/DataContainer.php
```diff
@@ -755,4 +774,17 @@ protected function generateGlobalButtons()
...
+
+	/**
+	 * Return the name of the current palette
+	 * @return string
+	 */
+	abstract public function getPalette();
+
+	/**
+	 * Save the current value
+	 * @param mixed
```
aschempp: Missing variable name (`@param mixed $varValue`)

system/modules/core/classes/Theme.php
```diff
@@ -641,7 +643,7 @@ protected function extractThemeFiles($arrFiles, $arrDbFields)
 	 */
 	public function exportTheme(\DataContainer $dc)
 	{
-		// Get the theme meta data
+		/** @var \DataContainer|object $dc */
```
aschempp: This should be a method property, and it should also not force a type then. That would make it easier to just pass in an object with property `id`.

system/modules/core/classes/Theme.php
```diff
@@ -882,7 +884,7 @@ protected function addTableTlImageSize(\DOMDocument $xml, \DOMElement $tables, \
 		$imageSizeItemTable->setAttribute('name', 'tl_image_size_item');
 		$imageSizeItemTable = $tables->appendChild($imageSizeItemTable);
 
-		// Get all sizes
+		/** @var \Database\Result|object $objTheme */
```
aschempp: Maybe we could use `@return \Database\Result|object` for the execute method?

system/modules/core/controllers/BackendMain.php
```diff
@@ -176,95 +260,4 @@ protected function welcomeScreen()
 
 		return $objTemplate->parse();
 	}
-
-
-	/**
-	 * Output the template file
-	 */
-	protected function output()
```
aschempp: This is not a good idea, I'm pretty sure there are people using `BackendMain` as there parent class to override stuff like that.

system/modules/core/elements/ContentAlias.php
```diff
@@ -39,7 +39,7 @@ public function generate()
 			return '';
 		}
 
-		// Overwrite the ID
+		/** @var \ContentElement $objElement */
```
aschempp: Why is this necessary? `\ContentModel::findByPk` should already know the correct type?

system/modules/core/elements/ContentDownloads.php
```diff
@@ -21,7 +21,7 @@ class ContentDownloads extends \ContentElement
 
 	/**
 	 * Files object
-	 * @var \FilesModel
```
aschempp: You could actually use `@var \FilesModel[]` here and then `foreach` over the values. Then the IDE would already know the type of each item in the foreach-loop.

system/modules/core/forms/FormFileUpload.php
```diff
@@ -238,6 +250,8 @@ public function validate()
 
 					// Generate the DB entries
 					$strFile = $strUploadFolder . '/' . $file['name'];
+
+					/** @var \FilesModel $objFile */
```
aschempp: `\FilesModel::findByPath` should already return the correct type.

system/modules/core/library/Contao/Folder.php
```diff
@@ -339,6 +340,7 @@ protected function getHash()
 	{
 		$arrFiles = array();
 
+		/** @var \RecursiveDirectoryIterator $it */
```
aschempp: really?

system/modules/core/library/Contao/Image.php
```diff
@@ -869,20 +869,21 @@ public static function create($image, $size=null)
...
-					 ->setTargetHeight($imageSize->height)
-					 ->setResizeMode($imageSize->resizeMode)
-					 ->setZoomLevel($imageSize->zoom);
+			$imageObj->setTargetWidth($imageSize->width);
+			$imageObj->setTargetHeight($imageSize->height);
+			$imageObj->setResizeMode($imageSize->resizeMode);
+			$imageObj->setZoomLevel($imageSize->zoom);
 		}
 
+		/** @var \FilesModel $fileRecord */
```
aschempp: Should not be necessary

system/modules/core/library/Contao/Model.php
```diff
@@ -132,6 +134,8 @@ public function __construct(\Database\Result $objResult=null)
 			foreach ($arrRelated as $key=>$row)
 			{
 				$table = $this->arrRelations[$key]['table'];
+
+				/** @var static $strClass */
 				$strClass = static::getClassFromTable($table);
```
aschempp: `getClassFromTable` should already `@return static`

system/modules/core/library/Contao/Picture.php
```diff
@@ -118,6 +120,7 @@ public static function create($file, $size=null)
 			$picture->setImageSizeItems(\ImageSizeItemModel::findVisibleByPid($imageSize->id, array('order'=>'sorting ASC')));
 		}
 
+		/** @var \FilesModel $fileRecord */
```
aschempp: This should not be necessary

system/modules/core/library/Contao/String.php
```diff
@@ -579,6 +579,7 @@ public static function srcToInsertTag($data)
 				continue;
 			}
 
+			/** @var \FilesModel $file */
```
aschempp: This should not be necessary

system/modules/core/models/ContentModel.php
```diff
@@ -14,6 +14,333 @@
...
+ * @property integer $com_perPage
+ * @property boolean $com_moderate
+ * @property boolean $com_bbcode
+ * @property boolean $com_disableCaptcha
+ * @property boolean $com_requireLogin
+ * @property string  $com_template
+ * @property string  $typePrefix
+ * @property string  $classes
+ *
+ * @method static $this findById()
```
aschempp: `$this` must be wrong in this case.

system/modules/core/modules/Module.php
```diff
@@ -68,6 +188,7 @@ public function __construct($objModule, $strColumn='main')
 
 		parent::__construct();
 
+		/** @var \ModuleModel $objModule */
```
aschempp: Thes supported types should be defined in the method property.

system/modules/core/modules/ModuleBooknav.php
```diff
@@ -154,6 +155,7 @@ protected function getBookPages($intParentId, $groups, $time)
 		{
 			while ($objPages->next())
 			{
+				/** @var \PageModel $objPages */
```
aschempp: if `\PageModel::findPublishedSubpagesWithoutGuestsByPid` would `@return \Model\Collection|static[]` and we'd use `foreach` that would not be necessary.

leofeyer:
@aschempp Can you please stop posting "should not be necessary" comments? Every change I have made was necessary, so unless it works in your PhpStorm instance …

system/modules/core/modules/ModuleRssReader.php
```diff
@@ -153,6 +160,7 @@ protected function compile()
 		$items = array();
 		$last = min($limit, count($arrItems)) - 1;
 
+		/** @var \SimplePie_Item[] $arrItems $i */
```
aschempp: `$i` should not be here

system/modules/comments/classes/Comments.php
leofeyer: Nope.

system/modules/comments/classes/Comments.php
leofeyer: Yes, it is necessary. The method returns a model collection but the IDE needs to understand that we are accessing the model.

system/modules/core/classes/Ajax.php
leofeyer: True.

system/modules/core/pages/PageRegular.php
```diff
@@ -87,6 +89,7 @@ public function generate($objPage, $blnCheckRequest=false)
 			{
 				while ($objModules->next())
```
aschempp: If we use `foreach` on the collection, you could use `@var \ModuleModel[] $objModules`

system/modules/core/classes/Backend.php
leofeyer: True.

system/modules/core/classes/DataContainer.php
leofeyer: You are right. There are several of these wrong usages.

system/modules/core/widgets/FileSelector.php
```diff
@@ -200,9 +208,9 @@ protected function renderFiletree($path, $intMargin, $mount=false, $blnProtected
 		$xtnode = 'tree_' . $this->strTable . '_' . $this->strName;
 
 		// Get session data and toggle nodes
-		if ($this->Input->get($flag.'tg'))
+		if (Input::get($flag.'tg'))
```
aschempp: Missing global namespace prefix.

system/modules/core/classes/Theme.php
leofeyer: I'll change this.

system/modules/core/classes/Theme.php
leofeyer: I am not changing the type of `$objSizes` but of `$objTheme` :)

aschempp: True, sorry. I think the comment should be where you declare the variable then.

system/modules/core/controllers/BackendMain.php
leofeyer: I have reverted this already because it had other side effects, too.

system/modules/core/elements/ContentAlias.php
leofeyer: Yes, but `$objElement` is a content model first and then a content element. Ugly, I know :(

system/modules/core/forms/FormFileUpload.php
leofeyer: No, it does not, because `findByPath` returns a collection (the path is not unique). Stumbled upon it as well.

system/modules/core/library/Contao/Folder.php
leofeyer: Yep :)

There is another strange PHP behavior going on in the `Theme` class, which I however did not fix: "Method 'getAttribute' not found in class \DOMNode" :)

system/modules/core/library/Contao/Model.php
leofeyer: No, it actually returns a string.

system/modules/core/modules/Module.php
leofeyer: We'd still need to tell the IDE that we are using a `ModuleModel` here.

system/modules/core/modules/ModuleRssReader.php
leofeyer: True.

system/modules/core/pages/PageRegular.php
leofeyer: You are right and this probably applies to all places where we are traversing a model collection.

system/modules/core/elements/ContentAlias.php
aschempp: Aah, that happens later. Maybe move the comment to the place it actually happens…

system/modules/core/widgets/FileSelector.php
leofeyer: True.

system/modules/core/forms/FormFileUpload.php
aschempp: Hmm, could we use `@return \Model\Collection|static` for all collections? Because the API for would be identical and available?

leofeyer:
All issues mentioned fixed in 6df04c2b92de3b1f64e612d86916191dc2440f01.

system/modules/core/pages/PageRegular.php
aschempp: Jup. I stopped using `->next()` because it can lead to issues (due to the internal position pointer). As an example, if you use `getRelated` with a collection as result, each call would get you the same collection with the previous internal pointer (possibly set to the last result, so `next()` would return `false`).

system/modules/core/modules/Module.php
aschempp: Sorry I meant `@property \ModuleModel|\Model\Collection|\Database\Result $objModule`

system/modules/core/library/Contao/Model.php
aschempp: True, agreed.

leofeyer:
I am evaluating your proposal in `system/modules/core/pages/PageRegular.php`:

```php
// Create a mapper array in case a module is included more than once (see #4849)
if ($objModules !== null)
{
	/**  @var \ModuleModel[] $objModules */
	foreach ($objModules as $objModule)
	{
		$arrMapper[$objModule->id] = $objModule;
	}

	…
```

This is what you meant, right? There are about 200 places where it would have to be changed.

aschempp:
Yes that's what I mean. Great work so far! 

I dislike the `@method` for magic methods. I personally think people should use `findBy('field', 'value')`, there's no reason for the magic methods…

leofeyer:
I always use them :)

aschempp:
> I always use them :)

I know ^^

leofeyer:
While thinking about this, should we not generally replace every `@return Model\Collection` with `@return Model[]`?

aschempp:
That would mean the Collection methods (like `next` and `reset`) would not be available. But `@return \Model\Collection|static[]` should do it.

leofeyer:
Unfortunately it does not work like this:

```php
// Does not work
/**
 * @return static[]|\Model\Collection|null
 */

// This works though
/**
 * @return \FilesModel[]|\Model\Collection|null
 */
```

aschempp:
what happens if you just use `static`? because the model properties are also available on the model collection?

leofeyer:
I am done. Feel free to check the changes again before I merge the PR.

system/modules/core/drivers/DC_Table.php
```diff
@@ -15,99 +15,102 @@
  * Provide methods to modify the database.
  *
  * @author Leo Feyer <https://github.com/leofeyer>
+ * @author Andreas Schempp <https://github.com/schempp>
```
aschempp: Not sure if I contributed here, but my GitHub user is "aschempp" ;-)

system/modules/core/elements/ContentGallery.php
```diff
@@ -284,12 +286,15 @@ protected function compile()
 			// Do not index or cache the page if the page number is outside the range
 			if ($page < 1 || $page > max(ceil($total/$this->perPage), 1))
 			{
+				/** @var \PageModel $objPage */
```
aschempp: This global statement is actually unnecessary because `$objPage is already defined in the first line of `compile()` method.

system/modules/core/library/Contao/Folder.php
aschempp: Aah, got it. But using `@var \RecursiveDirectoryIterator[] $it` and a foreach-loop would be way more obvious ^^

system/modules/core/modules/ModuleRegistration.php
```diff
@@ -14,6 +14,8 @@
 /**
  * Front end module "registration".
  *
+ * @property array $editable
```
aschempp: This is unnecessary because parent class `Module` already defines the property

system/modules/core/drivers/DC_Table.php
leofeyer: Typo, I'm sorry.

system/modules/core/pages/PageError403.php
```diff
@@ -39,7 +39,7 @@ public function generate($pageId, $objRootPage=null)
 			$objRootPage = \PageModel::findPublishedById(is_integer($objRootPage) ? $objRootPage : $objRootPage->id);
 		}
 
-		// Look for an error_403 page
+		/** @var \PageModel $obj403 */
```
aschempp: Unnecessary, `find403ByPid` already returns `static`

system/modules/core/modules/ModuleRegistration.php
leofeyer: True.

system/modules/core/pages/PageError404.php
```diff
@@ -68,7 +69,7 @@ public function generate($pageId, $strDomain=null, $strHost=null, $blnUnusedGet=
 			}
 		}
 
-		// Look for an 404 page
+		/** @var \PageModel $obj404 */
```
aschempp: Also unnecessary

leofeyer:
The first two issues have been fixed in 0cb856664a133acdca4cd817087c45d6b3551f97. The third issue `\PageModel $obj403` is an issue I have come across quite often:

```php
$obj403 = \PageModel::find403ByPid($objRootPage->id);

// Die if there is no page at all
if ($obj403 === null)		// <-- PhpStorm recognizes the PageModel here
{
	header('HTTP/1.1 403 Forbidden');
	die_nicely('be_forbidden', 'Forbidden');
}

// Generate the error page
if (!$obj403->autoforward || !$obj403->jumpTo)		// <-- PhpStorm no longer recognizes the PageModel here
{
	…
```

This seems to be caused by the strict comparison operator (`===`). If you use the normal operator (`==`) or if you use a yoda condition (`null === $obj403`), everything works as expected. Adding another `/** @var … */` comment also fixes it.

Would you prefer the yoda condition over the comment?

system/modules/core/library/Contao/Folder.php
leofeyer: I'll change it.

aschempp:
I would got for yoda then, as that's anyway what we're gonna use in Contao 4

system/modules/core/pages/PageRegular.php
```diff
@@ -232,6 +238,7 @@ protected function getPageLayout($objPage)
 			die_nicely('be_no_layout', 'No layout specified');
 		}
 
+		/** @var \LayoutModel $objLayout */
```
aschempp: Yoda here too

system/modules/core/pages/PageRoot.php
```diff
@@ -38,9 +39,12 @@ public function generate($pageId, $blnReturn=false)
 			die_nicely('be_no_active', 'No active pages found');
 		}
 
+		/** @var \PageModel $objNextPage */
```
aschempp: Yoda problem again. I'll stop with that, guessing you'll find them ;-)

system/modules/news/modules/ModuleNewsArchive.php
```diff
@@ -161,12 +164,15 @@ protected function compile()
 				// Do not index or cache the page if the page number is outside the range
 				if ($page < 1 || $page > max(ceil($total/$this->perPage), 1))
 				{
+					/** @var \PageModel $objPage */
 					global $objPage;
```
aschempp: objPage is already set global at the beginning of the method.

system/modules/news/modules/ModuleNews.php
```diff
@@ -350,10 +369,12 @@ protected function generateNewsUrl($objItem, $blnAddArchive=false)
 
 	/**
 	 * Generate a link and return it as string
-	 * @param string
-	 * @param object
-	 * @param boolean
-	 * @param boolean
+	 *
+	 * @param string  $strLink
+	 * @param object  $objArticle
```
aschempp: `$objArticle` should be of type `\NewsModel`.

aschempp:
`Controller::printArticleAsPdf` should expect `\ModuleModel` as argument.

system/modules/core/controllers/FrontendIndex.php
```diff
@@ -56,7 +56,9 @@ public function __construct()
 	 */
 	public function run()
 	{
+		/** @var \PageModel $objPage */
```
aschempp: You should remove this as it prevents line 93 from correctly noticing the variable could be a `\Model\Collection`

leofeyer: Thanks a lot. I just stumbled across this, too, and I had not yet figured out :)

aschempp:
According to [PhpDoc](http://manual.phpdoc.org/HTMLSmartyConverter/PHP/phpDocumentor/tutorial_tags.property.pkg.html) we sould use `@property-read` and `@property-write` for read-only or write-only properties. For example this affects class `\Contao\Request`, where `code`, `error` etc. are only readable.
It does not make a difference for the IDE though.

Apart from that, I think we're done here. Need to iron some clothes now ^^

leofeyer:
Big thanks for taking the time!

leofeyer:
Merged in 5b8c493d0856bf6038f6e76b92225335a1635952.

PR #7644, opened 2015-02-18
Fixed problem when including the same form elements on the same page hav...

### LINKED ISSUE #7626: Form CSS ids sowie Element CSS ids nicht eindeutig, wenn das Form mehrfach verwendet wird
Reproduzierbar wie folgt:
* Form erstellen
* Mehrmals in die Seite einfügen

Nun hat das Form immer die gleiche CSS id, die Elemente ebenfalls. Das kommt z.B. recht häufig bei Responsive Webdesign vor, wenn z.B. Formulare an zwei verschiedenen Stellen eingebunden und dann dynamisch ein/ausgeblendet werden.

Ich habe mir den Code angeschaut und auf Anhieb keine einfache Lösung gefunden. Ansatz wäre die Modul/Artikel Hierachie in der ID abzubilden und z.B. die einzelnen IDs aneinander zu hängen.

Oder gibt es einen anderen Ansatz hierzu?

amenk:
...ing the same fix #7626 together with #7643

### DIFF
--- a/system/modules/core/library/Contao/Widget.php
+++ b/system/modules/core/library/Contao/Widget.php
@@ -123,6 +123,11 @@ abstract class Widget extends \BaseTemplate
 	 */
 	protected $objDca;
 
+	/**
+	 * Element usages during same request
+	 * @var array
+	 */
+	static protected $arrElementUsages = array();
 
 	/**
 	 * Initialize the object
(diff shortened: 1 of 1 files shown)


leofeyer:
What exactly is the initial problem here? Your fix is confusing, because it e.g. does not allow us to define the ID manually anymore.

leofeyer:
As discussed on Mumble on October 14th, this should be fixed in the `Form` class as suggested by @Toflar.

leofeyer:
Fixed in #7643.

PR #7646, opened 2015-02-18
Allow configurable less and scss import paths

grusch-it:
In some cases it would be helpful to be able to add additional import paths for especially sass.

Using configurable import paths it would be possible to use e.g. [zurbs foundation libsass template](https://github.com/zurb/foundation-libsass-template) out of the box. Currently you have to manually change [all imports](https://github.com/zurb/foundation-libsass-template/blob/master/scss/app.scss) for all the foundation components to relative ones, which is quite annoying at the moment.

With this solution anyone could add import paths using the `initconfig.php`.


leofeyer:
I wonder if the template can be complied with SASS 0.12?

grusch-it:
All foundation components do compile without any errors using [leafo/scssphp](https://github.com/leafo/scssphp) which is used by Contao (tested many times with 3.3.3 and 3.3.4).

leofeyer:
So, am I getting this right: to use the libsass template, you have to adjust two `@import` directives, but you prefer to adjust the `initconfig.php` file instead and add an import path to a global array, which is then added to the parser?

grusch-it:
Exactly. It's just nice to have and would make things a little bit easier.

Issue #7669, opened 2015-03-04
ActiveRecord in DataContainer does not handle virtual fields and load_callbacks


Toflar:
I found this issue when trying the "Update mode" in the "overrideAll" mode. I wanted to use the "add" option.

Example:
I have a list of channels whereas every channel belongs to one or more networks. So we have an `n:m` relation here.
When I edit a channel, I can choose from a checkbox widget, to what networks this channel belongs to. Now the problem is that this field called `networks` is a virtual field. It does not exist in the database because it is an `n:m` relation and values are loaded and stored using the `load_callback` and `save_callback` which works perfectly fine in the edit mode.

Now imagine I add a new network and I have 300 channels and I want them to be put in relation to that new network. The "overrideAll" mode is just perfect for that! Process:

* Click on "edit multiple"
* Choose 300 channels
* Click "Override"
* Choose the field "networks"
* Select the new network I added before and set the "Update mode" to "Add the selected values"

Perfect!
But: It doesn't work :-)

This is because of this piece of code here: https://github.com/contao/core/blob/master/system/modules/core/drivers/DC_Table.php#L2866-L2881

As you can see, it compares the new values to the old ones which are stored in `$this->objActiveRecord `. This however, does not work because in my case the value there is `null` as `$this->objActiveRecord` contains the values from the database only. You can see one occurrence here (there are multiple, unfortunately): https://github.com/contao/core/blob/master/system/modules/core/drivers/DC_Table.php#L2561-L2566

Two problems here:

1. Virtual fields are never handled
2. `load_callback`'s are never called

So not only is this a problem in my case where I have a virtual field but also it is an issue when you register a `load_callback` and want to adjust the values!

I think we should write a general `loadActiveRecord` method in `DataContainer` and replace all the places where the activeRecord is loaded by a call to `loadActiveRecord` (Note: Untested, should help to understand the problem and what I want to do):

```php
protected function loadActiveRecord($id)
{
    $arrFields = $GLOBALS['TL_DCA'][$this->strTable]['fields'];
    $objActiveRecord = new \stdClass();
    $objRow = $this->Database->prepare("SELECT * FROM " . $this->strTable . " WHERE id=?")
        ->limit(1)
        ->execute($id);

    // Fetch values and call load_callbacks
    foreach ($arrFields as $strField => $arrField)
    {
        $varValue = $objRow->{$strField};

        if (is_array($GLOBALS['TL_DCA'][$this->strTable]['fields'][$strField]['load_callback']))
        {
[...]

PR #7670, opened 2015-03-04
Fixed overrideAll mode

Toflar:
Fixed overrideAll mode not calling load_callback for comparison with old values.
Also see #7669

### DIFF
--- a/system/modules/core/drivers/DC_Table.php
+++ b/system/modules/core/drivers/DC_Table.php
@@ -2861,6 +2861,23 @@ class DC_Table extends \DataContainer implements \listable, \editable
 				$new = deserialize($varValue, true);
 				$old = deserialize($this->objActiveRecord->{$this->strField}, true);
 
+				// Call load_callback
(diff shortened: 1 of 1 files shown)


leofeyer:
Your fix is wrong. There are three cases:

**`save()` is called from `edit()`**

In this case `$this->activeRecord` contains the correct value already, which is set here:

https://github.com/contao/core/blob/master/system/modules/core/drivers/DC_Table.php#L1860-L1878

**`save()` is called from `editAll()`**

In this case `$this->activeRecord` contains the correct value, too, which is set here:

https://github.com/contao/core/blob/master/system/modules/core/drivers/DC_Table.php#L2279-L2297

**`save()` is called from `overrideAll()`**

In this case the load_callback is not triggered indeed. However, I don't know whether this is on purpose, because it is not possible (override all does not work on a particular row), or if we simply forgot.

leofeyer:
Sorry, my bad. Your PR is just fine :)

leofeyer:
Merged in c17e2e20f5aab9bf3e90bb82d37bf88a9cfb5606.

PR #7686, opened 2015-03-12
parseArticles hook as in ModuleNews

wtom:


leofeyer:
What do you need the hook for?

discordier:
I guess he wants to add custom data to each news item.
Something like adding support for a linked element or something like that comes into mind.

leofeyer:
@discordier So would you suggest to add the hook?

wtom:
discordier is right. also i think it's pretty good to have the same functionality in the different parts of contao. why having the hook in the news module and not in the article module.

discordier:
@leofeyer In the end it does not harm and I can think of valid reasons to need this hook. However I have not yet really needed it as I was always hacking away via template and the like.
I guess some people might find it handy.
Therefore +1 from me. What do the other @contao/developers think about it?

ausi:
Am I right that this PR suggests to use the same name for the new hook as the already existing parseArticles hook in [system/modules/news/modules/ModuleNews.php:187-195](https://github.com/contao/core/blob/1f7448f356cbef9299d7285b2b3068723e620eb1/system/modules/news/modules/ModuleNews.php#L187-L195)?

I don’t think this would be a good idea.

discordier:
Regarding the name, @ausi is right. We should name it different.

leofeyer:
@wtom Can you please adjust your PR accordingly? I'd suggest to name the hook "parseArticle" (singular).

ausi:
Wouldn’t `moduleArticleCompile`, `articleCompile` or `compileArticle` be a better name for this hook?

leofeyer:
I could live with `compileArticle`, too.

leofeyer:
Implemented in a9e21702c943751deb512e9de47a1687110f03f4.

PR #7694, opened 2015-03-16
Make newslist front end module customizable

Toflar:
So I had to build yet another filter module for the news list and yet another custom news list module replacing the whole `compile()` method because the one from the core cannot be extended at all and I thought this time I'm going to propose changes so it might possibly be added to 3.5 :)

Advantages:
- Easy to override the methods `countItems()` and `fetchItems()` in a child class
- Added hooks that allow certain modifications without even extending the module class

### DIFF
--- a/system/modules/news/modules/ModuleNewsList.php
+++ b/system/modules/news/modules/ModuleNewsList.php
@@ -91,7 +91,7 @@ class ModuleNewsList extends \ModuleNews
 		$this->Template->empty = $GLOBALS['TL_LANG']['MSC']['emptyList'];
 
 		// Get the total number of items
-		$intTotal = \NewsModel::countPublishedByPids($this->news_archives, $blnFeatured);
+		$intTotal = $this->countItems($this->news_archives, $blnFeatured);
 
 		if ($intTotal < 1)
 		{
@@ -147,11 +147,11 @@ class ModuleNewsList extends \ModuleNews
 		// Get the items
 		if (isset($limit))
 		{
-			$objArticles = \NewsModel::findPublishedByPids($this->news_archives, $blnFeatured, $limit, $offset);
+			$objArticles = $this->fetchItems($this->news_archives, $blnFeatured, $limit, $offset);
 		}
 		else
 		{
-			$objArticles = \NewsModel::findPublishedByPids($this->news_archives, $blnFeatured, 0, $offset);
+			$objArticles = $this->fetchItems($this->news_archives, $blnFeatured, 0, $offset);
 		}
 
 		// Add the articles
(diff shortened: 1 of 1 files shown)


qzminski:
+1 and introduce this in other modules :)

system/modules/news/modules/ModuleNewsList.php
```diff
@@ -162,4 +162,60 @@ protected function compile()
...
+	protected function countItems($newsArchives, $blnFeatured)
+	{
+		// HOOK: add custom logic
+		if (isset($GLOBALS['TL_HOOKS']['newsListCountItems']) && is_array($GLOBALS['TL_HOOKS']['newsListCountItems']))
+		{
+			foreach ($GLOBALS['TL_HOOKS']['newsListCountItems'] as $callback)
+			{
+				$intResult = \System::importStatic($callback[0])->$callback[1]($newsArchives, $blnFeatured, $this);
+
+				if ($intResult !== false && is_int($intResult))
```
leofeyer: Why are you checking for `!== false` if `$intResult` has to go through `is_int()`? Does `is_int()` return `true` if you pass `false` as argument?

system/modules/news/modules/ModuleNewsList.php
```diff
@@ -162,4 +162,60 @@ protected function compile()
...
+	protected function fetchItems($newsArchives, $blnFeatured, $limit, $offset)
+	{
+		// HOOK: add custom logic
+		if (isset($GLOBALS['TL_HOOKS']['newsListFetchItems']) && is_array($GLOBALS['TL_HOOKS']['newsListFetchItems']))
+		{
+			foreach ($GLOBALS['TL_HOOKS']['newsListFetchItems'] as $callback)
+			{
+				$objCollection = \System::importStatic($callback[0])->$callback[1]($newsArchives, $blnFeatured, $limit, $offset, $this);
+
+				if ($objCollection !== false && ($objCollection === null || $objCollection instanceof \Model\Collection))
```
leofeyer: Why are you checking for `!== false` if `$objCollection` has to be either `null` or an object instance?

system/modules/news/modules/ModuleNewsList.php
Toflar: Same question as @qzminski asked here: https://github.com/Toflar/contao-core-fork/commit/7266cf149d00710420a40f6d13970fbfc6afcb1d#commitcomment-10210935

system/modules/news/modules/ModuleNewsList.php
Toflar: Because `null` is a valid return type. There are 3 use cases:

1. I am responsible for this module and I return a collection
2. I am responsible for this module and I return `null` because nothing was found
3. I am not responsible

`false` is for case 3.

qzminski: Because if you return false then the hook will be skipped.

leofeyer:
Merged in 1fc0e1e7e8300617b570c0f148ee44243ac54881.

leofeyer:
I have also modified the hook routines to make it more obvious that returning `false` means "I am not responsible". Please review and let me know if it doesn't work for you this way.

Toflar:
Looks good to me, thx :)

qzminski:
Thanks for this! Can we introduce the same methods in other modues?

leofeyer:
Sure.

PR #7708, opened 2015-03-24
Fixed the navigation "trail" class for pages that are internal redirects

qzminski:


leofeyer:
What exactly does your code do?

qzminski:
Well currently if the page is a forward to another internal page, then this forward in navgation has an "active" class if we're on the jumpTo page. I have just added the "trail" class to this case. See example:

```
// Hidden pages
Page A
Page A.1

// Visible pages in the navigation
Page A forward
Page A.1 forward
```

If we are now on Page A.1, then the Page A.1 forward has class "active", but its parent (Page A forward) has no trail class. My commit fixes it.

@Toflar can you support me in explaining the problem if needed?

leofeyer:
How exactly do I need to configure the pages to reproduce this issue?

qzminski:
We have actually done it in one project with @Toflar so it would be the best to see it there, live. Here is a quick and primitive reproduction though:

2015-07-20 at 21 37

So if you would go to the ```Hidden pages > Page A.1``` the page ```Forward pages > Page A``` would not have the ```trail``` CSS class, although ```Forward pages > Page A.1``` would be ```active```.

qzminski:
Sorry for close/open, missed the button :blush:

leofeyer:
I actually cannot reproduce this. If I am on `Hidden pages > Page A.1`, the `Forward pages > Page A.1` page does **not** have the class `active`. It only has the class `forward`.

Toflar:
It's about the class `trail` not `active`?

leofeyer:
The `trail` class is assigned to all parent items of the `active` element. Therefore, if there is no child element with the `active` class, there must not be a `trail` class.

qzminski:
The ```active``` class still should be there. This behavior was tested in 3.4 so maybe in 3.5 have been changed. I will setup a demo online, test again and send it you.

qzminski:
Okay in Contao 3.4 the active forward item is ```<span>``` but it still does not have the ```active``` class. So this is yet another bug but the fix may be a BC break.

Regarding the ```trail``` class - @leofeyer I am sending you my private demo system via e-mail.

leofeyer:
No, it has a `<span>` element, because we are not allowed to add a self-link. It is not the active page and there cannot be two active pages, either. And there can only be one trail from the root page to the active page.

qzminski:
In a single navigation this case does not make sense, that's true. However, if you have two navigations - for example one in the header (regular pages) and one in the footer (forward pages) - it's impossible to check the active and trail page in the footer menu. If you need a real world example of this situation please contact @Toflar for URL.

PR #7710, opened 2015-03-24
Optimize the database queries

leofeyer:
This PR relates to #7450.

### DIFF
--- a/system/modules/comments/dca/tl_comments.php
+++ b/system/modules/comments/dca/tl_comments.php
@@ -30,9 +30,9 @@ $GLOBALS['TL_DCA']['tl_comments'] = array
 			'keys' => array
 			(
 				'id' => 'primary',
-				'source' => 'index',
-				'parent' => 'index'
-				// FIXME: combined index?
+				'source' => 'index', // FIXME: remove?
+				'parent' => 'index', // FIXME: remove?
+				'source,parent,published' => 'index'
 			)
 		)
 	),
--- a/system/modules/comments/dca/tl_comments.php
+++ b/system/modules/comments/dca/tl_comments.php
@@ -654,7 +654,7 @@ class tl_comments extends Backend
 		}
 
 		// Update the database
-		$this->Database->prepare("UPDATE tl_comments SET tstamp=". time() .", published='" . ($blnVisible ? 1 : '') . "' WHERE id=?")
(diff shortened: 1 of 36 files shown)


aschempp:
Would it be useful to move `$time = time() - (time() % 60);` to something like `\Date::minute()`?

leofeyer:
As discussed on Mumble on March, 26th, we should add a method `Date::floorToMinute()`:

```php
public static function floorToMinute($time = null)
{
    if ($time === null)
    {
        $time = time();
    }

    return $time - ($time % 60);
}
```

system/modules/comments/dca/tl_comments.php
```diff
@@ -30,8 +30,9 @@
 			'keys' => array
 			(
 				'id' => 'primary',
-				'source' => 'index',
-				'parent' => 'index'
+				'source' => 'index', // FIXME: remove?
+				'parent' => 'index', // FIXME: remove?
+				'source,parent,published' => 'index'
```
leofeyer: As discussed on Mumble on March, 26th, we should remove the `source` and `parent` field indexes and add a stand-alone index on the `published` field.

system/modules/comments/dca/tl_comments_notify.php
```diff
@@ -26,10 +26,11 @@
 			'keys' => array
 			(
 				'id' => 'primary',
-				'source' => 'index',
-				'parent' => 'index',
-				'tokenConfirm' => 'index',
-				'tokenRemove' => 'index'
+				'source' => 'index', // FIXME: remove?
+				'parent' => 'index', // FIXME: remove?
+				'tokenConfirm' => 'index', // FIXME: remove?
```
leofeyer: As discussed on Mumble on March, 26th, we should remove the single indexes on `source`, `parent` and `tokenConfirm`.

system/modules/core/dca/tl_page.php
```diff
@@ -43,9 +43,9 @@
 			'keys' => array
 			(
 				'id' => 'primary',
-				'pid' => 'index',
 				'alias' => 'index',
-				'type' => 'index'
+				'type' => 'index', // FIXME: remove?
```
leofeyer: As discussed on Mumble on March, 26th, the single index on the `type` column can be removed.

leofeyer:
Merged in 573b4b67a4d2c3a8f3fd28ace1a34392c1675ae6.

Issue #7724, opened 2015-03-26
Improve FilesModel::findByPath()


Toflar:
`FilesModel::findByPath()` always searches on the DB even though the model might have been added to the model registry before.

Toflar:
Closed in favor of https://github.com/contao/core/pull/7725

PR #7725, opened 2015-03-26
[RTM] Improve model registry

Toflar:
This is my PR to solve #7724 and even more!

Currently, our model registry only allows to register models by their PK which in general is `id`. This causes useless DB queries in a lot of circumstances. The one mentioned in #7724 was `tl_files.path` which is actually unique as well and currently when executing `FilesModel::findByPath('path')` 20 times, there will be 20 queries to the database even though it's always the same.

Note that this happens quite often since Contao 3.4, because our `Image` class does that for every image to check if an important part is set.

Now instead of just solving this issue I tried to find a general solution for those cases :-)
The model registry can now handle aliases that internally point to the ID again.
Let's say we have this in `tl_files`:

| id | path |
| --- | --- |
| 42 | path_to_my_file.jpg |

In Contao 3.4 **AND** my PR (= so it's fully backwards compatible) you can fetch the model from the registry via its ID:

``` php
Model\Registry::getInstance()->fetch('tl_files', 42);
```

With this PR you can now also do this:

``` php
Model\Registry::getInstance()->fetch('tl_files', 'path_to_my_file.jpg', 'path');
```

And thanks to the modifications on the `DcaExtractor` this **automatically** works for all fields that are unique either because of the SQL key definition:

``` php
'sql' => array
(
    'keys' => array
    (
        'uuid' => 'unique'
    )
)
```

or the `eval` section of the field itself:

``` php
'path' => array
(
    'eval'                    => array('unique'=>true),
    'sql'                     => "varchar(1022) NOT NULL default ''",
),
```

Every developer can either do that or override the `getUniqueFields()` method in their model class for full flexibility.

:tada:

### DIFF
--- a/system/modules/core/dca/tl_files.php
+++ b/system/modules/core/dca/tl_files.php
@@ -146,7 +146,8 @@ $GLOBALS['TL_DCA']['tl_files'] = array
 		),
 		'path' => array
 		(
-			'sql'                     => "varchar(1022) NOT NULL default ''"
+			'eval'                    => array('unique'=>true),
+			'sql'                     => "varchar(1022) NOT NULL default ''",
 		),
 		'extension' => array
 		(
--- a/system/modules/core/library/Contao/DcaExtractor.php
+++ b/system/modules/core/library/Contao/DcaExtractor.php
@@ -513,7 +513,7 @@ class DcaExtractor extends \Controller
 			{
 				$this->arrKeys[$field] = $type;
 
-				if ($type === 'unique')
+				if ($type == 'unique')
 				{
 					$this->arrUniqueFields[] = $field;
 				}
--- a/system/modules/core/library/Contao/DcaExtractor.php
+++ b/system/modules/core/library/Contao/DcaExtractor.php
@@ -69,6 +69,12 @@ class DcaExtractor extends \Controller
 	protected $arrOrderFields = array();
 
 	/**
+	 * Unique fields
+	 * @var array
+	 */
+	protected $arrUniqueFields = array();
+
+	/**
 	 * Keys
 	 * @var array
 	 */
@@ -205,6 +211,28 @@ class DcaExtractor extends \Controller
 	public function hasOrderFields()
 	{
 		return !empty($this->arrOrderFields);
+	}
+
+
+	/**
+	 * Return an array of unique columns
+	 *
+	 * @return array
+	 */
+	public function getUniqueFields()
+	{
+		return $this->arrUniqueFields;
+	}
+
+
+	/**
+	 * Return true if there are unique fields
+	 *
+	 * @return boolean True if there are unique fields
+	 */
+	public function hasUniqueFields()
+	{
+		return !empty($this->arrUniqueFields);
 	}
 
 
@@ -468,6 +496,11 @@ class DcaExtractor extends \Controller
 				{
 					$this->arrOrderFields[] = $config['eval']['orderField'];
 				}
+
+				if (isset($config['eval']['unique']) && $config['eval']['unique'] === true)
+				{
+					$this->arrUniqueFields[] = $field;
+				}
 			}
 		}
 
@@ -479,6 +512,11 @@ class DcaExtractor extends \Controller
 			foreach ($sql['keys'] as $field=>$type)
 			{
 				$this->arrKeys[$field] = $type;
+
+				if ($type === 'unique')
+				{
+					$this->arrUniqueFields[] = $field;
+				}
 			}
 		}
 
@@ -498,6 +536,7 @@ class DcaExtractor extends \Controller
 			}
 		}
 
+		$this->arrUniqueFields = array_unique($this->arrUniqueFields);
 		$this->blnIsDbTable = true;
 	}
 }
--- a/system/modules/core/library/Contao/Model.php
+++ b/system/modules/core/library/Contao/Model.php
@@ -66,6 +66,12 @@ abstract class Model
 	protected static $strPk = 'id';
 
 	/**
+	 * Class name cache
+	 * @var array
+	 */
+	protected static $arrClassNames = array();
+
+	/**
 	 * Data
 	 * @var array
 	 */
@@ -1140,9 +1146,15 @@ abstract class Model
 	 */
 	public static function getClassFromTable($strTable)
 	{
+		if (isset(static::$arrClassNames[$strTable]))
+		{
+			return static::$arrClassNames[$strTable];
+		}
+
 		if (isset($GLOBALS['TL_MODELS'][$strTable]))
 		{
-			return $GLOBALS['TL_MODELS'][$strTable]; // see 4796
+			static::$arrClassNames[$strTable] = $GLOBALS['TL_MODELS'][$strTable]; // see 4796
+			return static::$arrClassNames[$strTable];
 		}
 		else
 		{
@@ -1153,7 +1165,8 @@ abstract class Model
 				array_shift($arrChunks);
 			}
 
-			return implode('', array_map('ucfirst', $arrChunks)) . 'Model';
+			static::$arrClassNames[$strTable] = implode('', array_map('ucfirst', $arrChunks)) . 'Model';
+			return static::$arrClassNames[$strTable];
 		}
 	}
 
--- a/system/modules/core/library/Contao/Model.php
+++ b/system/modules/core/library/Contao/Model.php
@@ -273,12 +273,7 @@ abstract class Model
 	{
 		$objDca = \DcaExtractor::getInstance(static::getTable());
 
-		if ($objDca->hasUniqueFields())
-		{
-			return $objDca->getUniqueFields();
-		}
-
-		return array();
+		return $objDca->getUniqueFields();
 	}
 
 
--- a/system/modules/core/library/Contao/Model.php
+++ b/system/modules/core/library/Contao/Model.php
@@ -710,7 +710,7 @@ abstract class Model
 	public function onRegister(\Model\Registry $registry)
 	{
 		// Register aliases to unique fields
-		foreach ($this->getUniqueFields() as $strColumn)
+		foreach (static::getUniqueFields() as $strColumn)
 		{
 			$varAliasValue = $this->{$strColumn};
 			if (!$registry->isAliasRegistered($this, $strColumn, $varAliasValue))
@@ -729,7 +729,7 @@ abstract class Model
 	public function onUnregister(\Model\Registry $registry)
 	{
 		// Unregister aliases to unique fields
-		foreach ($this->getUniqueFields() as $strColumn)
+		foreach (static::getUniqueFields() as $strColumn)
 		{
 			$varAliasValue = $this->{$strColumn};
 			if ($registry->isAliasRegistered($this, $strColumn, $varAliasValue))
--- a/system/modules/core/library/Contao/Model.php
+++ b/system/modules/core/library/Contao/Model.php
@@ -703,6 +703,44 @@ abstract class Model
 
 
 	/**
+	 * Called when the model is attached/registered to the model registry
+	 *
+	 * @param \Model\Registry
+	 */
+	public function onRegister(\Model\Registry $registry)
+	{
+		// Register aliases to unique fields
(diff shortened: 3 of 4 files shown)


system/modules/core/dca/tl_files.php
```diff
@@ -146,7 +146,8 @@
 		),
 		'path' => array
 		(
-			'sql'                     => "varchar(1022) NOT NULL default ''"
+			'eval'                    => array('unique'=>true),
```
leofeyer: This is wrong, the path must not be unique.

leofeyer:
I have added the `[RFC]` tag to the title of the PR, assuming that it no longer is work in progress?

system/modules/core/dca/tl_files.php
Toflar: ??? That's what the whole PR and discussion today was about?? 20x `FilesModel::findByPath('path.jpg')` should return the same instance because it's unique? If not, what combination of columns does make it unique then?

leofeyer: Only the UUID is unique. If you e.g. use SyncCto it may happen that you have two or more entries of the same file with different UUIDs, therefore we decided not to make the path unique.

Toflar: "unique" in terms of "unique for this installation". Everywhere we use `findByPath()` we expect a model, not a collection (if it was not unique it would return a collection, right?).
See e.g.
https://github.com/contao/core/blob/develop/system/modules/core/controllers/BackendPopup.php#L96
https://github.com/contao/core/blob/develop/system/modules/core/forms/FormFileUpload.php#L253
https://github.com/contao/core/blob/develop/system/modules/core/library/Contao/Picture.php#L123
https://github.com/contao/core/blob/develop/system/modules/core/modules/ModuleRegistration.php#L431

leofeyer:
I really cannot merge it because of the `path` column not being unique. The discussion about it is here: https://github.com/contao/core/issues/5556

leofeyer:
After having read #5556 again, I'm no longer sure if the column is not unique after all. It seems we just weren't able to set a unique index on DB level.

Toflar:
My changes do *NOT* set any keys :) It's just our `eval->unique` feature which in DC_Table is checked here: https://github.com/contao/core/blob/master/system/modules/core/drivers/DC_Table.php#L2855
It's not even in use for `DC_File` so it won't have any effect at all when editing files in the back end. The only thing it does is, it makes sure that 20 `FilesModel::findByPath('foobar.jpg')` calls all return the same model and do not execute 20 queries.

leofeyer:
Ok, so we must **never** set a unique index on `tl_files.path` or `tl_files.pid` and `tl_files.path` combined. It will throw an exception if the file paths exceed 250 characters (as discussed in #5556 and verified again in a test right now). It does, however, work with a regular btree index.

This means that we can actually merge the PR, because the column is unique (we just cannot set an index on DB level).

Toflar:
Okay, but please do not merge it yet. I would like to improve some naming things and try if can support arbitrary aliases that do not necessarily have to match 1 column only (so in theory it would be possible to alias multiple columns).

leofeyer:
I won't. The title still says `[RFC]` and not `[RTM]` :)

system/modules/core/library/Contao/DcaExtractor.php
```diff
@@ -479,6 +512,11 @@ protected function createExtract()
 			foreach ($sql['keys'] as $field=>$type)
 			{
 				$this->arrKeys[$field] = $type;
+
+				if ($type === 'unique')
```
leofeyer: We are only using `==` in the legacy code.

Toflar:
So there's some more work done here but leaving on `RFC` because I want to explain what I did here.
Now every `Model` has the possibility to do things with the registry in the `onRegister()` and `onUnregister()` methods/callbacks. By default those ones add aliases for unique fields to the registry, so you don't have to do anything here if you want this feature. However, now you can optimize your own models to now execute the query again.
Let's say you have something like this in your model:

```php
public static function findOneByExpensiveFilterQueryMethod(array $criteria)
{
    // Build complex query
    return static::findOneBy($queryParts);
}
```

Now you still have the issue that you might call this method with **the exact same** `$criteria` array and thus the query gets executed 20 times although your models have actually been loaded before.

You can now optimize like this:

```php
public static function findOneByExpensiveFilterQueryMethod(array $criteria)
{
    $alias = 'my_superb_alias';
    // Of course you may need to order the criteria here or so etc.
    $aliasValue = md5($criteria);

    $model = \Model\Registry::getInstance()->fetch(static::getTable(), $aliasValue, $alias);

    if ($model !== null) {
        return $model;
    }

    // Build complex query
   
    $model = static::findOneBy($queryParts);

    if ($model !== null) {
        \Model\Registry::getInstance()->registerAlias($model, $alias, $aliasValue);
        return $model;
    }

    return null;
}
```

Or use the `onRegister()` and `onUnregister()` callbacks etc. Lots of possibilities, now :)

system/modules/core/library/Contao/Model.php
```diff
@@ -678,6 +703,44 @@ public function attach()
...
+	 *
+	 * @param \Model\Registry
+	 */
+	public function onRegister(\Model\Registry $registry)
+	{
+		// Register aliases to unique fields
+		foreach (static::getUniqueFields() as $strColumn)
+		{
+			$varAliasValue = $this->{$strColumn};
+			if (!$registry->isAliasRegistered($this, $strColumn, $varAliasValue))
```
leofeyer: Please rename this method to `isRegisteredAlias()`.

leofeyer:
I like it. Can I merge it now?

system/modules/core/library/Contao/Model.php
Toflar: Done in 3c7e12cbfecb34036ac4ff2f2744140f59185eff.

Toflar:
If none of the @contao/developers has any objections, yes :)

leofeyer:
I am in the process of releasing the beta version of Contao 3.5. How long shall I wait?

Toflar:
Okay for me if this gets a bit more time to be reviewed by the other devs and makes it into the RC later in April :)

leofeyer:
:+1:

system/modules/core/library/Contao/Model.php
```diff
@@ -258,6 +264,25 @@ public static function getPk()
...
+	 * Return an array of unique field/column names
+	 * Do not include the PK here as this is handled separately
+	 *
+	 * @return array
+	 */
+	public static function getUniqueFields()
+	{
+		$objDca = \DcaExtractor::getInstance(static::getTable());
+
+		if ($objDca->hasUniqueFields())
```
discordier: This can be omitted as `\DcaExtractor` is initializing to empty array.
Simply always return the value of `$objDca->getUniqueFields()`

Toflar: True, „fixed“ in 2337fa60114952936e7aae4a8155a608c0f67521.

system/modules/core/library/Contao/Model.php
```diff
@@ -891,13 +937,22 @@ public static function findOneBy($strColumn, $varValue, array $arrOptions=array(
 	 */
 	public static function findBy($strColumn, $varValue, array $arrOptions=array())
 	{
+		$blnModel = false;
+
+		$arrColumn = (array) $strColumn;
```
discordier: The naming of `$strColumn` is now misleading, so is the phpDoc.
This should be changed to reflect the new behaviour.

system/modules/core/library/Contao/Model/Registry.php
```diff
@@ -83,16 +89,49 @@ public function count()
 	/**
 	 * Fetch a model by table name and primary key
 	 *
-	 * @param string  $strTable The table name
-	 * @param integer $intPk    The primary key
+	 * @param string  $strTable     The table name
+	 * @param mixed   $varKey       The key
+	 * @param string  $strAlias     An optional alias
```
discordier: Why reindenting the params?
Should be only one line added instead of -2/+3.

system/modules/core/library/Contao/Model.php
Toflar: Why? I did not change this afaik?

system/modules/core/library/Contao/Model/Registry.php
Toflar: I found it's too close to the key name, not really a problem, is it? :D

system/modules/core/library/Contao/Model/Registry.php
```diff
@@ -83,16 +89,49 @@ public function count()
...
 	 *
 	 * @return \Model|null The model or null
 	 */
-	public function fetch($strTable, $intPk)
+	public function fetch($strTable, $varKey, $strAlias = null)
 	{
-		if (isset($this->arrRegistry[$strTable][$intPk]))
+		// Default is searching by PK and is the most common case
```
discordier: I know we are in legacy code here and not in Contao 4 but please please please let's not increase the complexity of the old classes if we can use methods for new code even if it is in legacy classes.

system/modules/core/library/Contao/Model/Registry.php
discordier: Merely cosmetic but I think we should have a standard here. Usually it is one space after the longest parameter...

discordier:
Added all of my remarks.

system/modules/core/library/Contao/Model/Registry.php
Toflar: Better in 81bd651afcee93a6165ba9f3700f635b68e33f8d?

system/modules/core/library/Contao/Model.php
discordier: I may be a string or an array in your implementation but was only a string before.
Otherwise the check on `count($arrColumn) == 1` is not needed as it will always be 1.

Toflar: No, it could always be an array.

discordier: Then the description was misleading from the beginning. :)

leofeyer:
@Toflar Is this RTM now?

Toflar:
Yes. :)

leofeyer:
Great!

leofeyer:
Can you please take a look at the following two FIXMEs?

https://github.com/contao/core/blob/ce8e8241861143abe545678c1bd98386f6ce3742/system/modules/core/library/Contao/Model.php#L1041

https://github.com/contao/core/blob/ce8e8241861143abe545678c1bd98386f6ce3742/system/modules/core/library/Contao/Model.php#L1053

Toflar:
To be honest, I don't know why we're using `mergeRow()` there because if a model can be found in the registry you want to have this one back not an instance merged with DB data? This would mean that it's possible that the model in the registry does not contain the same data as in the DB row which would render the whole registry obsolete :) So imho the first FIXME can be ignored because it's correct to return the model instance from the registry if it's available and the second FIXME is invalid in terms of it's correct to check the registry again because my new code is **before** the query and the old and **after** the query. So it might be that the new code cannot find a model in the registry without executing the query (if this does not match: https://github.com/contao/core/blob/ce8e8241861143abe545678c1bd98386f6ce3742/system/modules/core/library/Contao/Model.php#L1046) but after the query is being executed it's still possible that it returns a result that is already in the registry so it's not necessary to instantiate that object again. However that `mergeRow()` call there still looks wrong to me as this would modify a model in the registry.

leofeyer:
What about [line 1048](https://github.com/contao/core/blob/ce8e8241861143abe545678c1bd98386f6ce3742/system/modules/core/library/Contao/Model.php#L1048)? Is `$arrOptions['value']` or `$arrOptions['value'][0]` always `$intId`? Or should we rename `$intId` to `$varKey`?

Toflar:
`$varKey` would be more appropriate because while in most cases it's an integer (if you call `findByPk(42)`), it will be a string for e.g. the FilesModel if you call `findByPath('path/to/file.jpg')`, you're right.

leofeyer:
Merged in 9da066cc982f4704361287f9c5c3f57148ee5001 then :)

PR #7755, opened 2015-04-09
Reset new fields to empty on version restore

aschempp:
If a field is added after storing a version of a record, it's value will stay when restoring. That's incorrect as the version did not contain that value.

see https://github.com/isotope/core/issues/1408#issuecomment-90911149


leofeyer:
Merged in b33ce6becf67dd1f48214fdbac28120cee3f3ea8.

PR #7763, opened 2015-04-14
Check all table rows upon theme import, not only the first one

ausi:
Currently the `Theme::compareThemeFiles()` method only checks the column names of the first row.

### DIFF
--- a/system/modules/core/classes/Theme.php
+++ b/system/modules/core/classes/Theme.php
@@ -198,13 +198,25 @@ class Theme extends \Backend
 					continue;
 				}
 
-				$fields = $rows->item(0)->childNodes;
+				$arrFieldNames = array();
+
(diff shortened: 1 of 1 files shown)


leofeyer:
Why is this a problem? Shouldn't all rows of a table have the same field names?

ausi:
Because extensions can change the XML file in the `exportTheme` hook this may not always be consistent. I stumbled upon this issue because I removed all unnecessary fields (with e.g. empty or `NULL` values) from the XML-File to reduce the file size.

leofeyer:
So this is actually a fix for a problem that you have created yourself in your extension :laughing:

ausi:
I would say I “discovered” this bug with my extension ;)

leofeyer:
LOL

PR #7764, opened 2015-04-15
added support for reading symbolic file links

JanMalte:


leofeyer:
What's the initial problem that your PR solves?

JanMalte:
Having a release base deployment tool, you generally share the localconfig.php and some other files.
For the most files this PR is working, but the Config is still recreated as a new tmp file and renamed afterwards, which still removes the existing symlink by a regular file.

leofeyer:
Still not sure how to reproduce. Do I have to make `localconfig.php` a symlink pointing to another file? And then the symlink gets overwritten by the `Config` class instead of the target file?

JanMalte:
I'm using Maggalanes (http://magephp.com/) for deployment.
Therefor my localconfig.php ins placed in ```/srv/project/shared/system/config/localconfig.php``` instead of ```/srv/project/releases/20150501120000/system/config/localconfig.php```, which is just a symbolic link to the file in ```/srv/project/shared/system/config/localconfig.php```.

If the settings are now changed with Contao, the symbolic link gets replaced by a real file with the new settings. It would be nice, if the symbolic link wouldn't be replaced by a new file. Instead the target file of the symbolic link should be replaced or changed.

This PR is just a first step, as the "bug" is placed in the Config class, which creates a new temporary file first, and moves this temporary file into place at the very end of the call.
https://github.com/contao/core/blob/960155bf2c6225cdf2f3d2f21fbae9492f86a11/system/modules/core/library/Contao/Config.php#L259

The changes of this PR do not address the issue with the config files, but with other files loaded by Contao.

leofeyer:
I have just checked and the code changes will not solve your issue, because the `Config` class does not use the `Files` class:

https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Config.php#L247-L249

JanMalte:
Yes, the Config class is the next step. I'm right now looking into this issue. This existing PR just solves problems with other symbolic links. It would be nice, if this could be merged into develop soon.

I will create a new PR for the config file issue.

JanMalte:
I've created a new commit with the needed changes, to support symlinked config files.There is also a test file for this issue: https://gist.github.com/JanMalte/ed3d7f1a7d9ed07768fe

system/modules/core/library/Contao/Files/Php.php
```diff
@@ -57,6 +57,9 @@ public function rmdir($strDirectory)
 	 */
 	public function fopen($strFile, $strMode)
 	{
+		if (is_link($strFile)) {
+			$strFile = readlink($strFile);
```
leofeyer: Shouldn't you prepend `TL_ROOT . '/'` here as well?

JanMalte: ```TL_ROOT``` is prepended in the second next line, so prepending it here would cause double prepending.

I even think it should NOT be prepended anyhow, if the path is a symbolic link, as the target could be outside of ```TL_ROOT```.

leofeyer: Yes, please change it. And also ensure that the logic is the same in both the `fopen()` and the `rename()` method.

thomaspahl:
For a relative symlink (ln -s ../b/file file)  ``readlink`` also  returns only the relative value (says doc, verified on Ubuntu/PHP-5.5 and cygwin/PHP-5.6, not yet tested on Windows with native symlinks - Win7ff). Thus you must evaluate it to an absolute path in the context of the link.

Generally, I feel a bit uncomfortable with manually resolving a symlink and would rather leave the job to the OS. With ``fopen/fwrite`` this is not a problem at all, both ``fopen(..'w')`` and ``fopen(..'w+')`` write to the linked file and leave the link in proper order (verified as above). With some paranoia, one could use ``fopen(...'w+')`` for an existing symlink to make sure the old file is used, but is is not necessary.

So the minimal invasive operation would be, in my opinion, to change only the rename function to evaluate symlinks - or (if possible, e.g. for config) to use ``fopen/fwrite`` instead of ``mv`` in case of an existing symlink (I know that this is no longer atomic but I wonder if that will be an issue in normal operation).

From my active times, I remember some issues across mounted filesystems (both nfs and samba) - but we are on the same mount here most likely. Note that the problem is the same for hard links.

My test code:
``` php
<?php
// run test in separate subdir /var/www/test (e.g.) to avoid simplicity effects
// (cd /var/www;echo "hi there">bin/x.txt;ln -sf bin/x.txt x;cd test;php test.php)
$link_rel = '../x';
$link_abs = '/var/www/x';
$file_rel = readlink($link_rel);
$file_abs = readlink($link_abs);
printf("%s=>%s - %s=>%s\n", $link_rel, $file_rel, $link_abs, $file_abs);
// get absolute path
if ($file_abs[0] != '/') {
    // may not work with native Windows paths
    $file_abs = dirname($link_abs) . '/' . $file_abs;
    printf("result: %s\n", $file_abs);
}
// verify linked file is written, not the link
echo file_get_contents($file_abs);
$fp = fopen($link_rel, 'w');
fwrite($fp, "changed there\n");
fclose($fp);
if (is_link($link_rel)) {
    echo "still a link\n";
}
echo file_get_contents($file_abs);
```
output:
```
www-data@heimdall:~/local$ php x2.php
../x=>bin/x.txt - /var/www/x=>bin/x.txt
result: /var/www/bin/x.txt
hi there
still a link
changed there
```

system/modules/core/library/Contao/Files/Php.php
thomaspahl: @JanMalte: in your test case you only use absolute paths, if I'm correct. For a relative symlink (ln -s ../b/file file)  ``readlink`` also  returns only the relative value (says doc, verified on Ubuntu/PHP-5.5 and cygwin/PHP-5.6, not yet tested on Windows with native symlinks - Win7ff). Thus you must evaluate it to an absolute path in the context of the link.

Generally, I feel a bit uncomfortable with manually resolving a symlink and would rather leave the job to the OS when possible. With ``fopen/fwrite`` this is not a problem at all, both ``fopen(...'w')`` and ``fopen(...'w+')`` write to the linked file and leave the link in proper state (verified as above). With some paranoia, one could use ``fopen(...'w+')`` for an existing symlink to make sure the old file is used, but is is not necessary.

So the minimal invasive operation would be, in my opinion, to change only the rename function to evaluate symlinks - or (if possible, e.g. for config) to use ``fopen/fwrite`` instead of ``mv`` in case of an existing symlink (I know that this is no longer atomic but I wonder if that will be an issue in normal operation).

My minimal test code:
```
<?php
// run test in separate subdir /var/www/test (e.g.) to avoid simplicity effects
// (cd /var/www;echo "hi there">bin/x.txt;ln -sf bin/x.txt x;cd test;php test.php)
$link_rel = '../x';
$link_abs = '/var/www/x';
$file_rel = readlink($link_rel);
$file_abs = readlink($link_abs);
printf("%s=>%s - %s=>%s\n", $link_rel, $file_rel, $link_abs, $file_abs);
// get absolute path
if ($file_abs[0] != '/') {
    // may not work with native Windows paths...
    $file_abs = dirname($link_abs) . '/' . $file_abs;
    printf("abs result: %s\n", $file_abs);
}
// verify linked file is written, not the link
echo file_get_contents($file_abs);
$fp=fopen($link_rel, 'w');
fwrite($fp, "changed there\n");
fclose($fp);
if (is_link($link_rel)) {
    echo "still a link\n";
}
echo file_get_contents($file_abs);
```
The output:
```
www-data@heimdall:~/test$ php x2.php
../x=>bin/x.txt - /var/www/x=>bin/x.txt
abs result: /var/www/bin/x.txt
hi there
still a link
changed there
```

JanMalte:
I have added some more tests locally, as I don't know where to place the tests suite for the changes of Contao files. Is there any unit test repository?
The tests include absolute and relative tests for each method in ```system/modules/core/library/Contao/Files/Php.php ```

leofeyer:
@JanMalte Are `copy()` and `rename()` the only affected methods?

@contao/developers Please review and vote.

Toflar:
Not sure whether the File class should handle symlinks by default. Actually I doubt it should...

leofeyer:
@Toflar What are your concerns?

Toflar:
Magic? Well, I mean at least I should be able to tell the class whether it should follow symlinks or not, right? Otherwise, I have no control at all. But well, `file_get_contents()` and others follow them silently as well, so I don't know what's the right way. I just feel uncomfortable with the change, maybe for no reason.

leofeyer:
Symlinks is a very helpful and fundamental Unix feature, which we should support IMHO. Especially if PHP does, too.

Toflar:
Sure, didn't say I don't want them. I just said I don't know if it should be enabled by default. Probably yes, but somehow I should have the option to tell that I don't want to follow them.

leofeyer:
Fair enough.

discordier:
Am I the only one who thinks the behaviour implemented in this PR is inconsistent?
The move operation will now overwrite the target file and therefore follow the symlink and therefore keep it while the copy operation is now stabilizing the symlinks and therefore creating a hard copy of the file.
I think this is counter intuitive.

leofeyer:
As discussed on Mumble on August 13th, we are not happy with this PR, e.g. your code behaves ambiguously as pointed out by @discordier. We should preserve the symlink of course. And if the symlink shall be copied or renamed, we must not copy or rename the target file.

leofeyer:
I think it is best to just behave like PHP. What happens if you apply `copy()` to a symlink? Does it copy the symlink or the target?

leofeyer:
We have just tested it and PHP actually copies the target file. So the behavior of the `File` class is actually correct.

aschempp:
Unfortunately, the original problem is not fixed. I'm also using Magallanes, and the first change to the localconfig will create a new copy which means the file is no longer shared :-(

leofeyer:
We have discussed this in Mumble on January 26th and we have agreed that this would be an incompatible change of behavior, which could only be implemented in the next major version (Contao 5). However, the file class will no longer be used in Contao 5.

PR #7781, opened 2015-04-27
Allow to select multiple checkboxes while holding the SHIFT key

qzminski:
See the video for demo.

http://screencast.com/t/VeGQVlfRZ


aschempp:
+1000

Toflar:
+ 10 billion!!!

leofeyer:
Nice :)

qzminski:
Just please check the minified version before merging as my minify script made a one line out of the whole file, whereas there were two previously. Not sure where does this difference comes from.

Aybee:
Nice, thank you.

leofeyer:
Merged in 7b588637289dde5073841b3c07b07582fb79514e.

leofeyer:
Just found a bug: it only works when clicking the checkboxes directly. However, you can also toggle the checkboxes by clicking the rows. Can the script be adjusted to support this as well?

qzminski:
We have already discussed it with @aschempp and @Toflar. Can you guys elaborate on this?

aschempp:
We agreed thats not really necessary. To me, clicking the row is good if I need to click 100 rows because I would likely miss a checkbox. But that problem would be gone with this PR…

leofeyer:
Is it difficult to implement?

qzminski:
The one problem I see is that the current checkbox selection is made by the code in inline attribute ```onclick```.

Aybee:
Would be nice if you could solve this row click behavior. I'm working a lot in a 2.11 installation and there you have to hit this little checkbox. This is a pain. In Contao 3 this is much better.

And you not always have to choose a sequenced amount of checkboxes, sometimes you e.g. have to check every 3rd.

Can't you simply expand the label?

aschempp:
>And you not always have to choose a sequenced amount of checkboxes, sometimes you e.g. have to check every 3rd.

That's not related to this PR at all…

Aybee:
Hallo Andreas,

wenn ich das hier richtig verstanden habe, geht es doch darum, dass man mit Hilfe der Shift-Taste einen Bereich von Checkboxen anhaken kann anstatt jede Checkbox einzeln anzuhaken.

Jetzt habe ich da auch rausgelesen, dass dann die Option verloren geht, dass man eine Checkbox auch über die ganze Reihe anklicken kann. Wenn diese Option erhalten bleibt, dann kannst du meinen Beitrag dazu tatsächlich eintüten ;) Dann habe ich da was falsch verstanden und sofort Panik bekommen, dass wir jetzt die Checkboxen wieder genau treffen müssen.

leofeyer:
> dass wir jetzt die Checkboxen wieder genau treffen müssen

Doch, genau das ist der Fall, wenn Du das neue Feature nutzen möchtest.

leofeyer:
> The one problem I see is that the current checkbox selection is made by the code in inline attribute ```onclick```.

@qzminski This could be solved, I guess.

leofeyer:
Fixed and improved in a55aa68701a983bd4d3ab5cc66e07724f751a7a8. You can now click either the row or the checkbox while holding down the SHIFT key.

There is only one minor issue: when clicking rows, e.g. in the member list, the browser makes a text selection, which the script then removes again. However, there is a little flash of the selection before it is removed; maybe you guys know how to prevent this?

leofeyer:
Ah, just found `user-select: none;` :) Will try this tomorrow.

aschempp:
Preventing user selection sounds like a very bad idea. Looks like I would not be able to select something anymore to copy paste?

leofeyer:
Fixed in 1434b1a3559d61f45bc2d4e52df38cb847f2c5c4.

PR #7796, opened 2015-05-03
Reset the file pointer when truncating the file

qzminski:
Here is the issue:

``` php
$file = new \File('test.txt');
$file->write('foobar');
dump($file->getContent());
$file->truncate();
$file->write('foobaz');
dump($file->getContent());
```

produces:

```
string(6) "foobar"
string(12) "foobaz"
```

So even if the file is being truncated, the file pointer remains at the old place. Thus the next write produces an empty binary string in front of the file (which is length of the previous file content).

### DIFF
--- a/system/modules/core/library/Contao/File.php
+++ b/system/modules/core/library/Contao/File.php
@@ -401,7 +401,7 @@ class File extends \System
 
 
 	/**
-	 * Truncate the file
+	 * Truncate the file and reset file pointer
 	 *
 	 * @return boolean True if the operation was successful
 	 */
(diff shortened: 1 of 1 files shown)


system/modules/core/library/Contao/File.php
```diff
@@ -410,6 +410,7 @@ public function truncate()
 		if (is_resource($this->resFile))
 		{
 			ftruncate($this->resFile, 0);
+            rewind($this->resFile);
```
leofeyer: You should use tabs to indent lines in Contao 3 :)

qzminski: Fixed in the follow-up commit.

leofeyer:
Merged in 31ddcb40572403557f58222fa1897407f1bd1dc4.

PR #7812, opened 2015-05-07
Fix method indexFreeRequest

bytehead:
This PR fixes the method `\Environment::get('indexFreeRequest')`.

Before, the method did just nothing when `$strRequest` wasn't matching `index.php`:

``` php
protected static function indexFreeRequest()
    {
        $strRequest = static::get('request');

        if ($strRequest == 'index.php')
        {
            return '';
        }

        return $strRequest;
    }
```

### DIFF
--- a/system/modules/core/library/Contao/Environment.php
+++ b/system/modules/core/library/Contao/Environment.php
@@ -446,14 +446,7 @@ class Environment
 	 */
 	protected static function indexFreeRequest()
 	{
-		$strRequest = static::get('request');
-
-		if ($strRequest == 'index.php')
-		{
-			return '';
-		}
-
-		return $strRequest;
+		return preg_replace('/index.php(\/)?/', '', static::get('request'));
 	}
 
 


leofeyer:
I don't see an improvement here. You have replaced a cheap comparison with an expensive `preg_replace()` call. How is this better?

bytehead:
Ok, here are some examples:

```php
var_dump(\Environment::get('indexFreeRequest'))
``` 

Request | Result current version | Result fixed Version
------------ | ------------- | ------------
`http://localhost/contao/index.php` | `string '' (length=0)` | `string '' (length=0)`
`http://localhost/contao/index.php/page` | `string 'index.php/page' (length=14)` | `string 'page' (length=4)`
`http://localhost/contao/index.php/page/services` | `string 'index.php/page/services' (length=23)`  | `string 'page/services' (length=13)`

As you see, the current version does __not__ remove the `index.php` part, when there is actually more then `index.php` in the request.

aschempp:
I agree this is confusing, but it would completely alter the functionality of the method.
I think it is supposed to get you `http://example.com/` instead of `http://example.com/index.php`, but it should not return `http://example.com/page` if there is an index in the URL (because that would not work without URL rewriting).

bytehead:
> I think it is supposed to get you http://example.com/ 

Definitely not, there's a method to get the URL: [Contao/Environment.php#L322](https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Environment.php#L322)

And if you read the doc block from the `request()` method:
```php
/**
 * Return the relativ path to the script and include the request (e.g. index.php?id=2)
 *
 * @return string The relative path to the script including the request string
 */
protected static function request()
``` 

it's totally clear that the request is without the protocol and domain parts.

As well the doc block from the `indexFreeRequest` method states clearly what should returned back:
```php
/**
 * Return the request string without the index.php fragment
 *
 * @return string The request string without the index.php fragment
 */
protected static function indexFreeRequest()

aschempp:
> definitely not, there's a method to get the URL: Contao/Environment.php#L322

Sorry I did not mean it should return the domain. Check out https://github.com/bytehead/core/commit/a2bef211d5c59604d6c1579d00ae2d28afb0601d to see where it is used. It's supposed to return a valid request URL to the current page, but not add `index.php` if you are on the root page.

bytehead:
Ok, I think I get your point. 

But the method is still totally wrong due to the description and my understanding of what this method should do: `* Return the request string without the index.php fragment`

aschempp:
I totally agree on that ^^

bytehead:
> (because that would not work without URL rewriting)

You're right, didn't recognize that.

leofeyer:
I agree that the method description is wrong. The method is supposed to return an empty string if the request is only `index.php` but not to strip the `index.php` fragment from each request string; the latter would not work as @aschempp explained.

aschempp:
@bytehead is there really a use-case why you need that method? Also, this will be solved in Symfony anyways…

bytehead:
No not really. I just used/saw it the first time and was a bit confused about the description and the actual functionality ^^

I can easily adapt this in my extension.

Symfony would be cool :sunglasses: but the project this time is still on 3.2.x

leofeyer:
Here's the original ticket: #2360 I still don't fully understand why we are only removing the `index.php` fragment if there is no request though :(

bytehead:
That's also my question :smile:

aschempp:
Because you don't want to redirect someone to `http://example.com/index.php`, because the `index.php` is unnecessary for the root.

bytehead:
So you could actually remove the `index.php` fragment as well on not-root-requests, if URL rewriting is enabled, don't you?

aschempp:
if URL rewriting is enabled, the request would not contain an `index.php` in the first place

Zeromax:
btw. should it remove "index.html" (file Extension from config) or "index/"?

aschempp:
@Zeromax not sure what you're referring to? This issue is only about `index.php`

Zeromax:
@aschempp I know. But I thought it should also strip index.html.
But if this method is only in use for index.php I am totally wrong ;)

PR #7834, opened 2015-05-22
simplify google analytics script

marcobiedermann:
Load Google Analytics script using async and defer. Discussed at HTML5 boilerplate https://github.com/h5bp/html5-boilerplate/pull/1696


leofeyer:
You cannot simply delete the first line. You have to take care that the `ga()` function is defined! Take a look at the ticket you have referenced; they have added the following:

```js
window.ga=function(){ga.q.push(arguments)};ga.q=[];ga.l=+new Date;
ga('create','UA-XXXXX-Y','auto');ga('send','pageview');
```

Also, make sure to read the [comment by @philipwalton](https://github.com/h5bp/html5-boilerplate/pull/1696#issuecomment-96118051) from the GA tracking team! He is suggesting another version which does not use the `defer` attribute.

Generally, I think it would be wise to wait until either the PR has been merged into HTML5 boilerplate or Google has officially announced a new snippet (or both).

marcobiedermann:
Sorry my bad, I forgot one line of code. Yeah, let's wait how they decide and pull this in if everything seems save

leofeyer:
We have meanwhile updated the script to the latest version suggested by Google.

Issue #7422, opened 2014-11-04
Small improvements 


alarstyle:
I have some suggestions about improvements.

1. Combine Page Picker and File Picker and make just one Picker. And it should work not only with pages and files, but with any table.
You just need to change the logic. Maybe it is possible to use regular view of the table, but instead of showing action icon you can show checkbox or radio.

2. Remove 
`onclick="Theme.toggleSelect(this)"`
Changing checkbox/radio state can be easily done with wrapping blocks into `label`s, and it will work without JS.

3. Remove 
`onmouseover="Theme.hoverDiv(this,1)"`
`onmouseout="Theme.hoverDiv(this,0)"`
`onmouseover="Theme.hoverRow(this,1)"`
`onmouseout="Theme.hoverRow(this,0)"`
It just changing background of elements, it can be done with CSS.

4. In the backend navigation, items always should be in DOM, even if the group is collapsed. It won't make the performance faster or slower, but it will prevent flashing of preloader.

5. Move all html code from drivers (DC_Table, DC_File, DC_Folder) to separate templates. So people can extend this templates.

6. Output website title from settings page in the header of backend. Something like this:
![screen_title](https://cloud.githubusercontent.com/assets/1730120/4914442/86b6f740-64c2-11e4-9c4c-7083af8b3df1.jpg)

7. When debug mode is on, in the backend there is inline style `margin-bottom: 60px;` for `<body>` tag. And it is changing when user toggles debug bar. It can be done with css, just need to add/remove css class in `<body>` tag. This can be useful, for example, if user wants to change look of debug bar or make it higher.

If something is not clear, just let me know and I'll try to explain :)

leofeyer:
We have discussed your ticket on Mumble on January 26th.

1. We don't think this is possible or worth the effort to make it possible.

2. This would require a markup change and thus imply a BC break.

3. This is something we can try.

4. Please make a pull request with the required changes. Make sure that the application correctly remembers the menu state.

5. Right now we cannot foresee the future of the `DC_*` drivers. It is possible that we will be incorporating the `DC_General` extension in one of the next major versions, then the existing drivers would be removed. Otherwise we would refactor this accordingly.

6. Please make a pull request with the required changes.

7. Please make a pull request with the required changes.

leofeyer:
See https://github.com/contao/core/issues/7837 regarding no. 3 and https://github.com/contao/core/issues/7838 regarding no. 4.

[...]

PR #7839, opened 2015-05-24
Issue #7422 no.7

alarstyle:
#7422 no.7

> When debug mode is on, in the backend there is inline style
> margin-bottom: 60px; for <body> tag. And it is changing when user
> toggles debug bar. It can be done with css, just need to add/remove css
> class in <body> tag. This can be useful, for example, if user wants to
> change look of debug bar or make it higher.


aschempp:
Is this a bug fix? I'm just asking because the debug bar has been removed in Contao 4 and 3.5 is already feature-complete.

alarstyle:
It's not realy a bug fix, it's more like feature that adds more flexibility.
I don't know about Contao 4, but as I can see, in 3.5 debug bar works the same as in 3.4.

aschempp:
The PR is :+1: from me. Dunno if it can still make it into 3.5 though because of the release state.

Ping @leofeyer

leofeyer:
> Dunno if it can still make it into 3.5 though because of the release state.

It cannot. Contao 3.5 is feature complete (RC1), so the ticket would have to be assigned to the next feature release. However, the next feature release is Contao 4.1 and we are no longer using the debug bar in Contao 4 :(

alarstyle:
It's a pity. In any case I was doing this pull request for too long, sorry for that.

alarstyle:
It's a pity. In any case I was doing this pull request for too long, sorry for that.

leofeyer:
I'm sorry it developed that way :(

leofeyer:
I just noticed that the PR does in fact improve the speed of the debug bar, because it no longer requires to change styles via JavaScript (like `$$('#contao-debug>*').setStyle('width',window.getSize().x);`). So we could see this as a bug fix I guess …

leofeyer:
Merged in 782a18b28b606e305bc8ef5b53f1cf2c65966c57.

PR #7852, opened 2015-06-02
[RTM] Sanitize the file names of uploaded files

leofeyer:
This PR will sanitize the file names of uploaded files and make sure that they are not exceeding the length of the database field.

### DIFF
--- a/system/modules/core/classes/FileUpload.php
+++ b/system/modules/core/classes/FileUpload.php
@@ -103,13 +103,19 @@ class FileUpload extends \Backend
 
 		foreach ($arrFiles as $file)
 		{
-			// Romanize the filename
-			$file['name'] = strip_tags($file['name']);
-			$file['name'] = utf8_romanize($file['name']);
-			$file['name'] = str_replace('"', '', $file['name']);
+			// Sanitize the filename
+			$file['name'] = \String::sanitizeFileName($file['name']);
+
+			// Invalid file name
+			if (!\Validator::isValidFileName($file['name']))
+			{
+				\Message::addError(sprintf($GLOBALS['TL_LANG']['ERR']['filename'], $file['name']));
+				$this->log('File "'.$file['name'].'" contains invalid characters or is longer than 255 characters' , __METHOD__, TL_ERROR);
+				$this->blnHasError = true;
+			}
 
 			// File was not uploaded
-			if (!is_uploaded_file($file['tmp_name']))
+			elseif (!is_uploaded_file($file['tmp_name']))
 			{
 				if ($file['error'] == 1 || $file['error'] == 2)
 				{
--- a/system/modules/core/classes/FileUpload.php
+++ b/system/modules/core/classes/FileUpload.php
@@ -110,8 +110,7 @@ class FileUpload extends \Backend
 			}
 			catch (\InvalidArgumentException $e)
 			{
-				\Message::addError(sprintf($GLOBALS['TL_LANG']['ERR']['filename'], $file['name']));
-				$this->log('File "'.$file['name'].'" cannot be sanitized' , __METHOD__, TL_ERROR);
+				\Message::addError($GLOBALS['TL_LANG']['ERR']['filename']);
 				$this->blnHasError = true;
 
 				continue;
@@ -120,8 +119,7 @@ class FileUpload extends \Backend
 			// Invalid file name
 			if (!\Validator::isValidFileName($file['name']))
 			{
-				\Message::addError(sprintf($GLOBALS['TL_LANG']['ERR']['filename'], $file['name']));
-				$this->log('File "'.$file['name'].'" is longer than 255 characters' , __METHOD__, TL_ERROR);
+				\Message::addError($GLOBALS['TL_LANG']['ERR']['filename']);
 				$this->blnHasError = true;
 			}
 
--- a/system/modules/core/classes/FileUpload.php
+++ b/system/modules/core/classes/FileUpload.php
@@ -104,13 +104,22 @@ class FileUpload extends \Backend
 		foreach ($arrFiles as $file)
 		{
 			// Sanitize the filename
-			$file['name'] = \String::sanitizeFileName($file['name']);
+			try
+			{
+				$file['name'] = \String::sanitizeFileName($file['name']);
+			}
+			catch (\InvalidArgumentException $e)
+			{
+				$this->log('File "'.$file['name'].'" cannot be sanitized' , __METHOD__, TL_ERROR);
+				$this->blnHasError = true;
+				continue;
+			}
 
 			// Invalid file name
 			if (!\Validator::isValidFileName($file['name']))
 			{
 				\Message::addError(sprintf($GLOBALS['TL_LANG']['ERR']['filename'], $file['name']));
-				$this->log('File "'.$file['name'].'" contains invalid characters or is longer than 255 characters' , __METHOD__, TL_ERROR);
+				$this->log('File "'.$file['name'].'" is longer than 255 characters' , __METHOD__, TL_ERROR);
 				$this->blnHasError = true;
 			}
 
--- a/system/modules/core/classes/FileUpload.php
+++ b/system/modules/core/classes/FileUpload.php
@@ -110,8 +110,10 @@ class FileUpload extends \Backend
 			}
 			catch (\InvalidArgumentException $e)
 			{
+				\Message::addError(sprintf($GLOBALS['TL_LANG']['ERR']['filename'], $file['name']));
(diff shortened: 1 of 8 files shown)


Toflar:
Looks good to me.

ausi:
Should we handle uploads from the frontend and backend the same way?

I still vote for a whitelist like `[^A-Za-z0-9_.-]`, at least for the frontend.

leofeyer:
> Should we handle uploads from the frontend and backend the same way?

Yes, why not? Wouldn't it be confusing if we didn't?

> I still vote for a whitelist like `[^A-Za-z0-9_.-]`, at least for the frontend.

You suggested `/(?:["\'<>]|[^\pL\pN\pP\pS])+/u`, right? That does not allow the "Mark" group (characters intended to be combined with another characters, e.g. accents, umlauts, enclosing boxes, etc.). Did you omit this on purpose?

Actually, I do have a general problem with the `utf8_romanize()` call. It seems that Contao as well as the browsers can handle UTF-8 characters, therefore I wonder why we are romanizing at all?

leofeyer:
Also, I'm not sure if we should forbid brackets and quotes by default. What about the following file name:

`reports "cleaned" < 2015.xls`

Isn't this a perfectly valid file name?

ausi:
> You suggested `/(?:["\'<>]|[^\pL\pN\pP\pS])+/u`, right?

Only if we really don’t want to use `[^A-Za-z0-9_.-]`.

> Actually, I do have a general problem with the `utf8_romanize()` call. It seems that Contao as well as the browsers can handle UTF-8 characters, therefore I wonder why we are romanizing at all?

IMO `utf8_romanize()` would only make sense if we use `[^A-Za-z0-9_.-]`.

>`reports "cleaned" < 2015.xls`
> Isn't this a perfectly valid file name?

It depends on the operating system and the file system. `<` and `"` are not allowed on Windows for example:

windows-8-filename

IMO it’s perfectly fine to rename the file to `reports-cleaned-2015.xls` if it was uploaded via the frontend.

ausi:
Maybe it’s better to discuss this issue in a Mumble call?

leofeyer:
We'll have our dev call tomorrow at 4:30pm. Do you have time?

ausi:
Yes.

system/modules/core/classes/FileUpload.php
```diff
@@ -103,13 +103,28 @@ public function uploadTo($strTarget)
...
-			$file['name'] = utf8_romanize($file['name']);
-			$file['name'] = str_replace('"', '', $file['name']);
+			// Sanitize the filename
+			try
+			{
+				$file['name'] = \String::sanitizeFileName($file['name']);
+			}
+			catch (\InvalidArgumentException $e)
+			{
+				$this->log('File "'.$file['name'].'" cannot be sanitized' , __METHOD__, TL_ERROR);
```
Toflar: So the end user does not get any message?

system/modules/core/forms/FormFileUpload.php
```diff
@@ -108,8 +108,27 @@ public function validate()
...
-		// Romanize the filename
-		$file['name'] = utf8_romanize($file['name']);
+		// Sanitize the filename
+		try
+		{
+			$file['name'] = \String::sanitizeFileName($file['name']);
+		}
+		catch (\InvalidArgumentException $e)
+		{
+			$this->addError(sprintf($GLOBALS['TL_LANG']['ERR']['filename'], $file['name']));
```
Toflar: Not sure if this is a good idea because it would output the invalid name which might contain anything bad!

system/modules/core/forms/FormFileUpload.php
```diff
@@ -108,8 +108,27 @@ public function validate()
...
+			$this->log('File "'.$file['name'].'" cannot be sanitized' , __METHOD__, TL_ERROR);
+
+			return;
+		}
+
+		// Invalid file name
+		if (!\Validator::isValidFileName($file['name']))
+		{
+			$this->addError(sprintf($GLOBALS['TL_LANG']['ERR']['filename'], $file['name']));
```
Toflar: Same here

system/modules/core/library/Contao/Validator.php
```diff
@@ -388,4 +388,35 @@ public static function isInsecurePath($strPath)
...
+		}
+
+		// Invisible control characters or unused code points
+		if (preg_match('/[\pC]/u', $strName))
+		{
+			return false;
+		}
+
+		// Must not be longer than 255 characters
+		if (utf8_strlen($strName) > 255)
```
Toflar: Imho, this should be the first check because it's likely the fastest of them all :)

system/modules/core/classes/FileUpload.php
discordier: IMO it should read:
```php
\Message::addError($GLOBALS['TL_LANG']['ERR']['invalid_filename']);
$this->log('File "'.$file['name'].'" cannot be sanitized' , __METHOD__, TL_ERROR);
$this->blnHasError = true;
```
With the corresponding language string added.

system/modules/core/library/Contao/Validator.php
ausi: IMO the preg_match in the first `if` is faster than `utf8_strlen()`

system/modules/core/forms/FormFileUpload.php
discordier: As I already said on mumble before, do NOT output the filename as it is broken and the resulting message is unpredictable.

system/modules/core/library/Contao/Validator.php
discordier: But it might choke on invalid UTF chars which are checked in 409.

ausi:
We should also escape [system/modules/core/templates/backend/be_popup.html5:52](https://github.com/contao/core/blob/8f678aaf08102432f943241a90d13c1c308741bd/system/modules/core/templates/backend/be_popup.html5#L52).

system/modules/core/dca/tl_files.php
```diff
@@ -621,7 +621,7 @@ public function showFile($row, $href, $label, $title, $icon, $attributes)
 		}
 		else
 		{
-			return '<a href="contao/popup.php?src=' . base64_encode($row['id']) . '" title="'.specialchars($title).'"'.$attributes.' onclick="Backend.openModalIframe({\'width\':'.$row['popupWidth'].',\'title\':\''.str_replace("'", "\\'", $row['fileNameEncoded']).'\',\'url\':this.href,\'height\':'.$row['popupHeight'].'});return false">'.Image::getHtml($icon, $label).'</a> ';
+			return '<a href="contao/popup.php?src=' . base64_encode($row['id']) . '" title="'.htmlspecialchars($title, ENT_COMPAT, Config::get('characterSet')).'"'.$attributes.' onclick="Backend.openModalIframe({\'width\':'.$row['popupWidth'].',\'title\':\''.str_replace("'", "\\'", $row['fileNameEncoded']).'\',\'url\':this.href,\'height\':'.$row['popupHeight'].'});return false">'.Image::getHtml($icon, $label).'</a> ';
```
ausi: `$row['fileNameEncoded']` should also get encoded in this line.

system/modules/core/library/Contao/Validator.php
ausi: 409 doesn’t check if it is valid UTF-8, in fact it can be bypassed with invalid UTF-8:

```php
preg_match('/[\pC]/u', "normal string") === 0
preg_match('/[\pC]/u', "string with null byte \0") === 1
preg_match('/[\pC]/u', "string with null byte \0 and invalid UTF-8 \xC3") === false
```

So we should change the check to `if (preg_match('/[\pC]/u', $strName) !== 0)`.

system/modules/core/forms/FormFileUpload.php
leofeyer: So do you want us not to tell the user which of their files is broken?

system/modules/core/dca/tl_files.php
leofeyer: `$row['fileNameEncoded']` is encoded already, isn't it?

leofeyer: No, you are right. I have changed it accordingly.

leofeyer:
I have adjusted everything according to your comments.

system/modules/core/library/Contao/Validator.php
Toflar: I think this test should be extended to

```php
// Must not be longer than 255 characters but at least 1 character
$length = utf8_strlen($strName);
if ($length > 255 || $length < 1)
```

Otherwise an empty string would be a valid file name.

leofeyer:
Merged in 6aaa26c5051b23b7881b4d73bd816aab567284d6.

PR #7856, opened 2015-06-02
Respect the entity settings in the textarea form field

qzminski:


leofeyer:
What is this good for? We are not modifying the return values of the other widgets, either.

https://github.com/contao/core/blob/master/system/modules/core/forms/FormTextField.php#L105-L115

qzminski:
Today I created a custom form (not with the form generator) with a textarea that allows HTML tags, precisely said it was a tinyMCE instance. The returned value by the widget is parsed with ```specialchars()``` which automatically encoded all tags. I don't think that behavior should happen if we have the ```allowHtml``` and ```preserveTags``` flags set.

Interesting thing is that the real value ```$varValue``` in the widget is absolutely correct! It's just that ```_get()``` method which causes the trouble. And I cannot fetch the value the other way than that :-(

qzminski:
Back to your point - exactly, you are not modifying it in other widgets, but for the textarea you do. That's the problem.

leofeyer:
I see. I don't know if we can change this in Contao 3.5 though, because it is not backwards compatible.

@contao/developers What do you think?

aschempp:
Well what would happen if we change it? Most people will not have `preserveTags` enabled. If they did, they probably manually decoded the values. That should still work, right?

qzminski:
@aschempp that would still work in custom forms, that's true. However, I think it was introduced to meet some form generator expectations and here it may be a BC break.

discordier:
While I agree on the fact that this is very bad behaviour of a form widget, I am afraid that it is a breaking change.

Any anontent which is "already encoded" will not get double encoded anymore and therefore might open up some security hole then.
I mean, currently the content is always encoded and you have to explicitely decode it in your code to get the original value. If we change this, the content will not be encoded anymore but still will get decoded in the custom extensions and therefore maybe create bad data.

Example:
`<script>alert('hello');</script>` will currently return: `&lt;script&gt;alert('hello');&lt;/script&gt;` and the coder has to explicitely decode to get the original HTML and therefore (currently correctly assumes) his code is safe.
If we change this however, we will have the original data passed which might contain malicious values as if I pass the encoded value `&lt;script&gt;alert('hello');&lt;/script&gt;` it will not be detected by the Input class AND/OR allowed html tags checks as it is not an html tag yet. The data WILL get converted by the extension code then as it is decoding the values (as it is assuming they have been encoded by the textarea).

Therefore we MUST NOT alter this behaviour as bugfix but only as new feature within the next minor.

Toflar:
I second @discordier's remarks.

Issue #7826, opened 2015-05-17
Contao loads desktop cache file if no mobile cache file is present


fritzmg:
### Problem description

If you use the server cache and a mobile page layout and there is a cache file present for the desktop version for a specific page, but not for the mobile version, then Contao will load the desktop version for mobile devices as well.

### Reproduction

* Activate the page cache
* Use different page layouts for desktop and mobile
* In your desktop browser, open a specific page in the frontend (without being logged into the backend of course) other than the index page (see #7618[#issuecomment-83171643](https://github.com/contao/core/issues/7618#issuecomment-83171643)), so that its cache file will be generated
* Now open the same page on your smartphone - it will load the desktop version

### Cause

The problem is the mobile check in [FrontendIndex.php::outputFromCache()](https://github.com/contao/core/blob/master/system/modules/core/controllers/FrontendIndex.php#L318):
```php
// Check for a mobile layout
if (\Input::cookie('TL_VIEW') == 'mobile' || (\Environment::get('agent')->mobile && \Input::cookie('TL_VIEW') != 'desktop'))
{
	$strMd5CacheKey = md5($strCacheKey . '.mobile');
	$strCacheFile = TL_ROOT . '/system/cache/html/' . substr($strMd5CacheKey, 0, 1) . '/' . $strMd5CacheKey . '.html';
	if (file_exists($strCacheFile))
	{
		$blnFound = true;
	}
}
// Check for a regular layout
if (!$blnFound)
{
	$strMd5CacheKey = md5($strCacheKey);
	$strCacheFile = TL_ROOT . '/system/cache/html/' . substr($strMd5CacheKey, 0, 1) . '/' . $strMd5CacheKey . '.html';
	if (file_exists($strCacheFile))
	{
		$blnFound = true;
	}
}
```
The first condition will set `$blnFound` to true, if there is a cache file present for the mobile version. However, if no cache file was found for the mobile version, the second condition will execute and thus the desktop cache file will be loaded, if there is one present, even if the requested page has a mobile layout defined.

I don't know of an easy fix, since we probably do not want to query whether the requested page needs to have a mobile layout or not.

Reference: https://community.contao.org/de/showthread.php?57144-Contao-l%E4dt-Desktop-Version-aus-Cache

dmolineus:
[...]

PR #7859, opened 2015-06-03
[RTM] Add a .desktop suffix to the cache file of a page with a mobile layout

leofeyer:
See #7826.


leofeyer:
I have updated the PR accordingly. Does it work now?

dmolineus:
Yes, it works with that changes.

leofeyer:
Fixed in 6aa701b67888106bb8fd003d061582601a4ab6e9 then.

PR #7860, opened 2015-06-03
[RTM] CSV values were not correctly loaded in editAll mode

aschempp:


leofeyer:
In the `edit()` method, the CSV fields are converted **before** the `load_callback`. You PR now adds the conversion **after** the `load_callback`. Can I assume that it should be the same in both methods?

aschempp:
Uhm, my PR is adding it before the callback?

leofeyer:
Hm, I need a vacation. Sorry :)

PR #7872, opened 2015-06-08
Update Frontenduser.php

stefanheimes:
There is a problem with ajax request in the frontend. An ajax request is added to the referer, but only the fe user has this problem. The be user has a check for this. See https://github.com/contao/core/blob/master/system/modules/core/classes/BackendUser.php#L116 and I find this commmit f3112ff4cc8cb114c72eab6f924f260f9cff4a20 which fix the same problem in the driver.

So I added a check for ajax request. Based on the BackendUser.php and the commit f3112ff4cc8cb114c72eab6f924f260f9cff4a20.

### DIFF
--- a/system/modules/core/classes/FrontendUser.php
+++ b/system/modules/core/classes/FrontendUser.php
@@ -73,7 +73,7 @@ class FrontendUser extends \User
 	{
 		$session = $this->Session->getData();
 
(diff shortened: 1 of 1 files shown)


leofeyer:
Merged in 63081812ee11228cdcefdaa2b8f0ddea2a4866ff.

PR #7882, opened 2015-06-12
Support dimensionless SVG images

ausi:
There are three types of SVG images regarding dimensions:
1. Fixed dimensions via `width` and `height` in `em`, `ex`, `px`, `pt`, `pc`, `cm`, `mm`, `in` or unitless but not percentage-based.
2. A defined `viewBox` but no or a percentage-based `width` and `height`.
3. No `viewBox` and no or a percentage-based `width` and `height`.

This is how the browsers handles them if you use ``:
1. Shows the image in the specified dimensions.
2. Resizes the image to 100% width and preserves the aspect ratio.
3. Displays the image in the dimensions 300x150

IMO `Image->imageSize` should only work for the first type. Resizing should work for type 1 and 2 because we have a defined aspect ratio. Resizing type 3 should just return the original file.
## What’s in this pull request

`Image->imageSize` now only works for GD images and SVG images of type 1.

Added `Image->imageViewSize` which returns the dimensions of `imageSize` if available or the dimensions of `viewBox` otherwise.

Changed image resizing to work with type 1 and 2 and doesn’t resize type 3.

Updated `Image::getPixelValue()` to `16px = 1em = 2ex = 12pt = 1pc = 1/6in = 2.54/6cm = 25.4/6mm = 100%` but didn’t remove the `%` because this would be a BC break, right? We should remove it in 4.1.0 though.

Updated file manager to work with type 2 and 3.

/cc @toflar

### DIFF
--- a/system/modules/core/classes/DataContainer.php
+++ b/system/modules/core/classes/DataContainer.php
@@ -511,7 +511,7 @@ abstract class DataContainer extends \Backend
 
 				if ($objFile->isSvgImage || $objFile->height <= \Config::get('gdMaxImgHeight') && $objFile->width <= \Config::get('gdMaxImgWidth'))
 				{
-					if ($objFile->width > 699 || $objFile->height > 524)
+					if ($objFile->width > 699 || $objFile->height > 524 || !$objFile->width || !$objFile->height)
 					{
 						$image = \Image::get($objFile->path, 699, 524, 'box');
 					}
@@ -525,7 +525,7 @@ abstract class DataContainer extends \Backend
 
 				$strPreview = '
 
-<div id="' . $ctrl . '" class="tl_edit_preview" data-original-width="' . $objFile->width . '" data-original-height="' . $objFile->height . '">
+<div id="' . $ctrl . '" class="tl_edit_preview" data-original-width="' . $objFile->viewWidth . '" data-original-height="' . $objFile->viewHeight . '">
 ' . \Image::getHtml($image) . '
 </div>';
 
--- a/system/modules/core/drivers/DC_Folder.php
+++ b/system/modules/core/drivers/DC_Folder.php
@@ -2395,8 +2395,17 @@ class DC_Folder extends \DataContainer implements \listable, \editable
 			// Generate the thumbnail
 			if ($objFile->isImage && $objFile->viewHeight > 0)
 			{
-				$popupWidth = ($objFile->width > 600) ? ($objFile->width + 61) : 661;
-				$popupHeight = ($objFile->height + 200);
+				if ($objFile->width && $objFile->height)
+				{
+					$popupWidth = ($objFile->width > 600) ? ($objFile->width + 61) : 661;
+					$popupHeight = ($objFile->height + 210);
+				}
+				else
+				{
+					$popupWidth = 661;
+					$popupHeight = 625 / $objFile->viewWidth * $objFile->viewHeight + 210;
+				}
+
 				$thumbnail .= ' <span class="tl_gray">('.$this->getReadableSize($objFile->filesize);
 				if ($objFile->width && $objFile->height)
 				{
@@ -2413,6 +2422,10 @@ class DC_Folder extends \DataContainer implements \listable, \editable
 			}
 			else
 			{
+				if ($objFile->isImage)
+				{
+					$popupHeight = 360;
+				}
 				$thumbnail .= ' <span class="tl_gray">('.$this->getReadableSize($objFile->filesize).')</span>';
 			}
 
--- a/system/modules/core/drivers/DC_Folder.php
+++ b/system/modules/core/drivers/DC_Folder.php
@@ -2393,18 +2393,22 @@ class DC_Folder extends \DataContainer implements \listable, \editable
 			$return .= "\n  " . '<li class="tl_file click2edit toggle_select" onmouseover="Theme.hoverDiv(this,1)" onmouseout="Theme.hoverDiv(this,0)"><div class="tl_left" style="padding-left:'.($intMargin + $intSpacing).'px">';
 
 			// Generate the thumbnail
-			if ($objFile->isImage && $objFile->height > 0)
+			if ($objFile->isImage && $objFile->viewHeight > 0)
 			{
 				$popupWidth = ($objFile->width > 600) ? ($objFile->width + 61) : 661;
 				$popupHeight = ($objFile->height + 200);
-				$thumbnail .= ' <span class="tl_gray">('.$this->getReadableSize($objFile->filesize).', '.$objFile->width.'x'.$objFile->height.' px)</span>';
+				$thumbnail .= ' <span class="tl_gray">('.$this->getReadableSize($objFile->filesize);
+				if ($objFile->width && $objFile->height)
+				{
+					$thumbnail .= ', '.$objFile->width.'x'.$objFile->height.' px';
+				}
+				$thumbnail .= ')</span>';
 
 				if (\Config::get('thumbnails') && ($objFile->isSvgImage || $objFile->height <= \Config::get('gdMaxImgHeight') && $objFile->width <= \Config::get('gdMaxImgWidth')))
 				{
-					$_height = ($objFile->height < 50) ? $objFile->height : 50;
-					$_width = (($objFile->width * $_height / $objFile->height) > 400) ? 90 : '';
+					$_height = ($objFile->height && $objFile->height < 50) ? $objFile->height : 50;
 
-					$thumbnail .= '<br><img src="' . TL_FILES_URL . \Image::get($currentEncoded, $_width, $_height) . '" alt="" style="margin:0 0 2px -19px">';
+					$thumbnail .= '<br><img src="' . TL_FILES_URL . \Image::get($currentEncoded, 400, $_height, 'box') . '" alt="" style="margin:0 0 2px -19px">';
 				}
 			}
 			else
--- a/system/modules/core/library/Contao/File.php
+++ b/system/modules/core/library/Contao/File.php
@@ -272,22 +272,12 @@ class File extends \System
 
 						$svgElement = $doc->documentElement;
 
-						if ($svgElement->getAttribute('width') && $svgElement->getAttribute('height'))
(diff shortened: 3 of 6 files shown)


Toflar:
Excellent work! :+1:  Regarding 

> but didn’t remove the % because this would be a BC break, right? We should remove it in 4.1.0 though.

1. If it was a BC break, we cannot change it in 4.1 either, we'd have for 5.0 as we use Semantic Versioning now. But
2. I think this is a bug because I cannot imagine anybody using this method with a percentage value to receive a useful value, right? It was just complete nonsense. So it can be fixed in 3.5.1 as well as 4.0.1 I guess.

ausi:
IMO keeping the `%` doesn’t hurt, I think we should add a deprecation message and remove it in 5.0.

Toflar:
I take your point.

Toflar:
Just had that problem again, it's definitely a bug because it's confusing people :)

ausi:
It is. In the last Contao call someone (I forgot who) talked about the same issue, so it’s not only you :)

aschempp:
Jup, we discussed that topic last week in the Mumble call

ausi:
@aschempp Do you remember who it was? Maybe he should try if this solution forks for him.

aschempp:
Dunno exactly, maybe it was @fiedsch?

fiedsch:
No, I wasn't. And sadly the only thing I recall is "there was somebody talking about that".

leofeyer:
As discussed on Mumble on August 13th, we want to merge this bug fix into Contao 3.5.3.

system/modules/core/drivers/DC_Folder.php
```diff
@@ -2393,18 +2393,22 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=f
...
+				$thumbnail .= ')</span>';
 
 				if (\Config::get('thumbnails') && ($objFile->isSvgImage || $objFile->height <= \Config::get('gdMaxImgHeight') && $objFile->width <= \Config::get('gdMaxImgWidth')))
 				{
-					$_height = ($objFile->height < 50) ? $objFile->height : 50;
-					$_width = (($objFile->width * $_height / $objFile->height) > 400) ? 90 : '';
+					$_height = ($objFile->height && $objFile->height < 50) ? $objFile->height : 50;
 
-					$thumbnail .= '<br><img src="' . TL_FILES_URL . \Image::get($currentEncoded, $_width, $_height) . '" alt="" style="margin:0 0 2px -19px">';
+					$thumbnail .= '<br><img src="' . TL_FILES_URL . \Image::get($currentEncoded, 400, $_height, 'box') . '" alt="" style="margin:0 0 2px -19px">';
```
leofeyer: Why did you replace `$_width` with `400` here?

system/modules/core/library/Contao/File.php
```diff
@@ -280,16 +280,6 @@ public function __get($strKey)
 								\Image::getPixelValue($svgElement->getAttribute('height'))
 							);
 						}
-						elseif ($svgElement->getAttribute('viewBox'))
```
leofeyer: Why did you remove the else condition here? If there is no `width` and `height` attribute, there might still be a `viewBox` attribute.

system/modules/core/widgets/FileSelector.php
```diff
@@ -359,16 +359,18 @@ protected function renderFiletree($path, $intMargin, $mount=false, $blnProtected
...
+					}
 
 					if (\Config::get('thumbnails') && ($objFile->isSvgImage || $objFile->height <= \Config::get('gdMaxImgHeight') && $objFile->width <= \Config::get('gdMaxImgWidth')))
 					{
-						$_height = ($objFile->height < 50) ? $objFile->height : 50;
-						$_width = (($objFile->width * $_height / $objFile->height) > 400) ? 90 : '';
+						$_height = ($objFile->height && $objFile->height < 50) ? $objFile->height : 50;
 
-						$thumbnail .= '<br><img src="' . TL_FILES_URL . \Image::get($currentEncoded, $_width, $_height) . '" alt="" style="margin:0 0 2px -19px">';
+						$thumbnail .= '<br><img src="' . TL_FILES_URL . \Image::get($currentEncoded, 400, $_height, 'box') . '" alt="" style="margin:0 0 2px -19px">';
```
leofeyer: Again, where does the hard-coded `400` come from?

system/modules/core/drivers/DC_Folder.php
ausi: I had to change the calculation of `$_width` and `$_height` because `$objFile->width` and `$objFile->height` can be `null` now. I simplified the calculation by using the `box` resize mode. `400` was hard coded before on [line 2405](https://github.com/madeyourday/contao-core/blob/83843f0ecb50c6d3b668859107b3416ed19dd0af/system/modules/core/drivers/DC_Folder.php#L2405).

system/modules/core/library/Contao/File.php
ausi: I explained it in https://github.com/contao/core/pull/7882#issue-87808347 `imageSize` must only respect fixed dimensions, not percentage or viewBox based values. I added `imageViewSize` which takes the `viewBox` into account if `imageSize` is not defined.

system/modules/core/widgets/FileSelector.php
ausi: See https://github.com/contao/core/pull/7882#discussion_r37399412

system/modules/core/classes/DataContainer.php
```diff
@@ -511,7 +511,7 @@ protected function row($strPalette=null)
 
 				if ($objFile->isSvgImage || $objFile->height <= \Config::get('gdMaxImgHeight') && $objFile->width <= \Config::get('gdMaxImgWidth'))
 				{
-					if ($objFile->width > 699 || $objFile->height > 524)
+					if ($objFile->width > 699 || $objFile->height > 524 || !$objFile->width || !$objFile->height)
```
leofeyer: Shouldn't this be `!$objFile->width && !$objFile->height`?

system/modules/core/drivers/DC_Folder.php
leofeyer: That's not quite correct. The maximum height is 50 but the width is only set to `90` if the aspect ratio is > 400. After your change, the width is always limited to 400, which it never was before.

ausi: Previously, images with an aspect ratio smaller than 400:50 (which applies to almost all images) were only limited for the height to 50 and because of the required aspect ratio they could never reach a width larger than 400. Images with an aspect ratio larger than 400:50 were cropped to the width of 90.

IMO the old behavior is a bit confusing, because a 400x50 image would be shown in 400x50 but a 401x50 image would be cropped to 90x50.

leofeyer: What about a 900x10 pixel image?

ausi: Previously this was cropped to 90x10. Now it would be resized to 400x4.

leofeyer: And do you expect anyone to recognize the image content if it is 400x4?

ausi: Yes, because all content of the image is still there, it’s just smaller. The 90x10 image has lost 90% of its contents because it was cropped.

system/modules/core/classes/DataContainer.php
ausi: If one of the two dimensions or both are missing we should resize the image, so IMO `!$objFile->width || !$objFile->height` is correct.

system/modules/core/library/Contao/File.php
```diff
@@ -316,6 +306,59 @@ public function __get($strKey)
...
+
+						$svgElement = $doc->documentElement;
+
+						if ($svgElement->getAttribute('viewBox'))
+						{
+							$svgViewBox = preg_split('/[\s,]+/', $svgElement->getAttribute('viewBox'));
+
+							$this->arrImageViewSize = array
+							(
+								intval($svgViewBox[2]),
```
leofeyer: Why are you no longer using `Image::getPixelValue()` here?

ausi: The `viewBox` attribute only allows unitless numbers, so `Image::getPixelValue()` should not be used here.

leofeyer:
I have found more problems while testing the changes.

**1. No dimensions are shown**



Shouldn't we fall back to showing the viewbox dimensions here?

**2. The info popup needs dimensions, too**



Since the image does not have dimensions, the popup window does not open high enough.

leofeyer:
No. 1 (no dimensions are shown) might be a non-issue. If we would show the viewbox dimensions, we could no longer determine that it is a dimensionsless image.

ausi:
> Shouldn't we fall back to showing the viewbox dimensions here?

I don’t think so, the image doesn’t have dimensions so we shouldn’t show them to the user. We could display an aspect ratio of the image, but I’m not sure in which format `4:3` or `1,33` or `100:75`? And I’m not sure if the user needs this information.

> Since the image does not have dimensions, the popup window does not open high enough.

I will look into that.

ausi:
The popup height should be fixed now.

system/modules/core/drivers/DC_Folder.php
```diff
@@ -2393,22 +2393,39 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=f
...
 				}
 			}
 			else
 			{
+				if ($objFile->isImage)
```
leofeyer: We get here if there is an image with a view height of less than `0`. When does this happen?

ausi: Less than or equal to `0` which is true for SVG images of type 3 described in https://github.com/contao/core/pull/7882#issue-87808347
Should I restructure the if/else to separate the generation of `$thumbnail` and `$popupWidth/$popupHeight`?

leofeyer: Why did you choose `360`?

ausi: Browsers display SVG images without dimension at 300x150. 360 is high enough that no scrollbar is needed in the popup, tested in latest Firefox on Mac.

leofeyer: That's probably because it follows the `$popupHeight = $objFile->height + 210` approach :)

Ok, I'm going to merge this now. Thanks for your patience.

leofeyer:
Merged in 6edc292a70cfaace4f9e505337d3f0ff41565dfe.

PR #7918, opened 2015-07-13
[RFC] Update countries.php

### LINKED ISSUE #7915: Add country for "Sint Maarten"
The country "Sint Maarten" is missing in the country list (country code "sx": https://en.wikipedia.org/wiki/ISO_3166-2:SX).

Do not confuse with Saint Martin (country code "mf": https://en.wikipedia.org/wiki/ISO_3166-2:MF).

Maybe we can update the whole country list with latest ISO 3166 data?

aschempp:
Sync missing countries and spellling with https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2

Fixes #7915


system/config/countries.php
```diff
@@ -56,7 +57,7 @@
 	'cf' => 'Central African Republic',
 	'cg' => 'Congo, Republic of',
 	'ch' => 'Switzerland',
-	'ci' => 'Cote d\'Ivoire',
+	'ci' => 'Côte d\'Ivoire',
```
discordier: AFAIR we only had lower ASCII chars in this file so far, the UTF-8 chars are noted in the language files only.
By changing this now, we are IMO introducing a breaking change here.

system/config/countries.php
```diff
@@ -215,14 +215,16 @@
...
-	'sk' => 'Slovak Republic',
+	'sk' => 'Slovakia',
 	'sl' => 'Sierra Leone',
 	'sm' => 'San Marino',
 	'sn' => 'Senegal',
 	'so' => 'Somalia',
 	'sr' => 'Suriname',
-	'st' => 'Sao Tome and Principe',
+	'ss' => 'South Sudan',
+	'st' => 'São Tomé and Príncipe',
```
discordier: Same as in https://github.com/contao/core/pull/7918/files#r34468434. We should stick to ASCII.

aschempp:
Not really true about the UTF8 (see https://github.com/contao/core/pull/7918/files#diff-88f8c38d4b99c62b78d59fa9b741f437R162 for example), but I still changed it. I also noticed we need to update the english language file.

@leofeyer maybe you can tell where the other translations came from? Because I don't know what we should use… The old files said `Reunion Island` but Wikipedia says `Réunion`. Obviously the old one is not wrong.

I changed the PR to RFC for now.

discordier:
@aschempp Then `Curaçao` appears to be the only UTF-8 notation in the whole file. Am I wrong?

aschempp:
I guess you're not, but I don't know what you wanna say with that :D

discordier:
I mean it is pretty strange to have all except for this one in ASCII. I simply wanted to state that it is awkward to me.

leofeyer:
As discussed on Mumble on July 16th, we should update the list according to https://www.iso.org/obp/ui/. Including non-ASCII characters.

aschempp:
Does that apply to the config file or the englisch translation?

leofeyer:
Updated in c4b3e431124952b221935bea3a4606a5ab4d7ec5.

aschempp:
Thanks! Though I think we should think more about users than technical correctness. All (the) do not add any relevant information. Same for UK…

Issue #4260, opened 2012-04-29
Add a hook to parseDate()


DyaGa:
First of all I have to say that it will be hard to explain what I mean, but I will try.

So, I just found an interesting i18n issue when parsing a date.

Example:

```php
<?php
echo $this->parseDate('d F Y');
```

The result will be (examples for languages):
```
en => 29 April 2012
ru => 29 Апрель 2012
uk => 29 Квітень 2012
```

The problem is that in Ukraine or Russia we don't say `29 Апрель 2012` or `29 Квітень 2012`, we say (and the result should be):
```
ru => 29 Апреля 2012
uk => 29 Квітня 2012
```

As I see in PHP documentation about [date function](http://php.net/manual/en/function.date.php) there is no equivalent for this situation.

So, I propose you to add one more array for months (for example `MONTHS_DAY`) to language file.
Next example is for Russian language:

```php
<?php
$GLOBALS['TL_LANG']['MONTHS_DAY'][0] = 'Января';
$GLOBALS['TL_LANG']['MONTHS_DAY'][1] = 'Февраля';
$GLOBALS['TL_LANG']['MONTHS_DAY'][2] = 'Марта';
$GLOBALS['TL_LANG']['MONTHS_DAY'][3] = 'Апреля';
$GLOBALS['TL_LANG']['MONTHS_DAY'][4] = 'Мая';
$GLOBALS['TL_LANG']['MONTHS_DAY'][5] = 'Июня';
$GLOBALS['TL_LANG']['MONTHS_DAY'][6] = 'Июля';
$GLOBALS['TL_LANG']['MONTHS_DAY'][7] = 'Августа';
$GLOBALS['TL_LANG']['MONTHS_DAY'][8] = 'Сентября';
$GLOBALS['TL_LANG']['MONTHS_DAY'][9] = 'Октября';
$GLOBALS['TL_LANG']['MONTHS_DAY'][10] = 'Ноября';
$GLOBALS['TL_LANG']['MONTHS_DAY'][11] = 'Декабря';
```

And in [System->parseDate function](https://github.com/contao/core/blob/2.11.x/system/libraries/System.php#L451) make some improvements, like this:

```php
<?php
	static function parseDate($strFormat, $intTstamp=null)
	{
		$strModified = str_replace
		(
			array('l', 'D', 'd F', 'j F', 'F', 'M'),
			array('w::1', 'w::2', 'd n::5', 'j n::5', 'n::3', 'n::4'),
			$strFormat
		);
[...]

PR #7921, opened 2015-07-14
Fixing parseDate() hook for working properly anywhere #4260

hamidpeywasti:
@leofeyer, I solved parseDate() hook problem. As you know, when in datetime format, theres is none of "l, D, F and M", the hook do not work, because of "retrun" in line 618. I need and use this hook for parsing persian date, so this pull request is needed to be merged, please review and test it.


aschempp:
There's a lot of wrong code style in your PR which makes it hardly understandable. However, the intent of the PR seems to execute the hook even if `strpos($strDate, '::') === false` matches. I don't know where this check came from…? It was added in TYPOlight 2.7 without anymore information.

hamidabbaszadeh:
In this pull request, I wanted to modify only Date.php! but thers's my previus pull request. My github knowledge is not good.

Yes, my goal is that (executaion of hook even if ```php strpos($strDate, '::') === false ```) for parsing date in Persian (Iranian) calendar that 100% differ from Gregorian calendar.

By using this change, parsing date in any type will be possible, but in current version we cannot.

Anyway, my code currently work. And if there any wrong code style, please guide me.

hamidabbaszadeh:
@leofeyer Thanks, You did it in b53c69d (#4260), so this pull request is not needed.

Issue #7734, opened 2015-03-30
Minify the markup must not remove all spaces, but collapse instead


Ruudt:
Contao settings allow to "Minify the markup". This removes whitespaces between HTML elements. But that alters the structure of the document. Therefore Contao should collapse whitespaces instead of removing them altogether.

In the examples below decrease the width of the "Result" pane until the menu doesn't fit. Because the HTML changed the results are different between the two examples. You can see the difference instantly because the spaces are visible in the first example. Tested with Chrome and Firefox.

Example with whitespaces:
https://jsfiddle.net/nkqwbfyo/

Example without whitespaces:
https://jsfiddle.net/nkqwbfyo/1/

discordier:
Just also ran into a similar issue with the file picker.
Filename is: `foo<space><space>bar.jpg`
The picker produces then:
```html
<input value="foo<space>bar.jpg"
```
Note the missing second space.
We therefore must not strip consecutive spaces in html attributes as well.

beaniemonk:
I was about to submit a sort-of-related ticket when I saw this one, so I hope you don't mind me piggybacking off it instead...

This is actually one of my favorite things about the HTML minifier :) It makes inline-block grids/layouts a breeze, and avoids having to use hacky workarounds like these:
https://css-tricks.com/fighting-the-space-between-inline-block-elements/

However, the minifier also tries to preserve white-space between inline HTML tags (which I agree is good) by using the following regex pattern: ```'/>\n<(a|input|select|span)/' => '> <$1'```

The problem I run into with the above is, in addition to preserving the white-space between tags like ```<a>```, ```<abbr>``` and ```<audio>``` (which again, I agree is good), it also preserves them for tags that start with 'a' but are actually *block* elements -- namely ```<address>```, ```<article>``` and ```<aside>```. This results in inconsistent behavior, and I end up having to avoid these elements in my custom templates so those extra spaces don't break stuff. And that's a shame, because all 3 very valuable, semantic elements.

So I wanted to propose one of two things:

1. Adjust the snippet above to only match *inline* elements that start with 'a', for example: ```'/>\n<(a(\b|bbr|udio)|input|select|span)/' => '> <$1'```

2. Move $arrReplace -- used in Template::minifyHtml() -- to a global configuration array so that we can customize it to our needs. (Otherwise this change will BC break several of our sites if we update them).

[...]

PR #7927, opened 2015-07-17
Fix markup minification

ausi:
See #7734

### DIFF
--- a/system/modules/core/library/Contao/Template.php
+++ b/system/modules/core/library/Contao/Template.php
@@ -56,6 +56,29 @@ abstract class Template extends \BaseTemplate
 	 */
 	protected $arrData = array();
 
+	/**
+	 * Valid JavaScipt types, see <http://www.w3.org/TR/html5/scripting-1.html#scriptingLanguages>
+	 * @var array
+	 */
+	protected static $validJavaScriptTypes = array(
+		'application/ecmascript',
+		'application/javascript',
+		'application/x-ecmascript',
+		'application/x-javascript',
+		'text/ecmascript',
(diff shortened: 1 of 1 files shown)


leofeyer:
Can you elaborate on your changes?

ausi:
I added a comment in https://github.com/contao/core/issues/7734#issuecomment-122367057 just now.

leofeyer:
In your implementation, you are preserving the chunk if it does not have a known JavaScript type. This implies that there is no optimization at all:

```html
<div id="my-slider">
    <img src="image1.jpg">
    <script type="text/html">
        <img src="image2.jpg">
    </script>
</div>
```

However, the chunk could be optimized:

```html
<div id="my-slider">
<img src="image1.jpg">
<script type="text/html">
<img src="image2.jpg">
</script>
</div>
```

Is there a reason why we should **not** optimize the markup this way?

ausi:
We should only optimize known types, `text/html` might be a special case which we could also optimize, but types like `text/plain` or `text/markdown` must not be optimized. E.g. changing

```html
<script type="text/markdown">
# Headline

    Code block starting with 4 spaces
</script>
```

to

```html
<script type="text/markdown">
# Headline
Code block starting with 4 spaces
</script>
```

would be bad.

Should I update the pull request to include the `text/html` case?

leofeyer:
No, I have already merged it locally.

leofeyer:
Merged in c4e4e6f098922d6a98a489b96d5cdb591550f934. I have removed the CSS check, because currently there is no other type than `text/css`.

PR #7934, opened 2015-07-22
Make compileCommands Public

amenk:
This is such a useful function...


leofeyer:
Since Contao 3.5 is already feature complete and the Installer class has been deprecated in favor of the installation-bundle – what am I supposed to do with the ticket?

PR #7942, opened 2015-07-28
New Function in FilesModel to list Subfolders

christianromeni:
This would be very usefull if you want to make sure all log entries are stored to a file before cleaning the log.
Many companies have to keep log-files, but dont want the log table to get full.
I have created a log-rotation module, that does that. But the log could be purged before beeing saved to file...

Here is the sorce of my module (unfinished and a little bit insecure up to now..):
https://github.com/christianromeni/cLogRotate

Also, there is only the option to look for multiple files in a folder, but not for multiple folders. This would be a nice feature to look for subfolders. maybe this could be extended to be recursive as well..

### DIFF
--- a/system/modules/core/library/Contao/Automator.php
+++ b/system/modules/core/library/Contao/Automator.php
@@ -114,9 +114,9 @@ class Automator extends \System
 	public function purgeSystemLog()
 	{
 		// HOOK: call before purge
-		if (isset($GLOBALS['TL_HOOKS']['beforeLogPurge']) && is_array($GLOBALS['TL_HOOKS']['beforeLogPurge']))
+		if (isset($GLOBALS['TL_HOOKS']['purgeSystemLog']) && is_array($GLOBALS['TL_HOOKS']['purgeSystemLog']))
 		{
-			foreach ($GLOBALS['TL_HOOKS']['beforeLogPurge'] as $callback)
+			foreach ($GLOBALS['TL_HOOKS']['purgeSystemLog'] as $callback)
 			{
 				$this->import($callback[0]);
 				$this->$callback[0]->$callback[1]();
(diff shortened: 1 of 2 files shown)


discordier:
First off, please do not mix two separate features in one PR, this makes things harder for us to read and if we should decide against one feature, the PR has to be rejected causing the second feature to get rejected also.

Second, while I do not have anything against the HOOK in general, I think the naming should rather be `purgeSystemLog` and the retrieval of the database object should stick to its usage. Therefore the HOOK should IMO be called first.

Third, on your new method in the `FilesModel`, I fail to see a usecase but I guess you have one at hand?

christianromeni:
Ok, sorry, I'm new to the PR thing... Also I'm not sure why the branch is called patch-1 ... I'll try and fix that tomorrow as well as splitting the PR up.

There is no problem calling it different. I'll change that.

A usecase for finding subfolder in my case is: I have definied one folder (eg: /files/logs/) but I'm dynamically creating subfilders depending on the year (eg: /files/logs/2015/).
I would now like to list all files in the subfolders of the defined folder.

This could also be used for multi-dimensional gallerys.

aschempp:
First of all, this feature would be added to the next version of Contao 4, and then we should use events and not hooks.
Secondly, I wonder if we should use Monolog for the backend logging, too? I'm pretty sure there would be event options there…

christianromeni:
So there is no chance for it to be added in any Contao 3 release?

aschempp:
Contao 3.5 is the last version of Contao 3, and a bugfix-release cannot contain new features…

leofeyer:
If you remove the logging hook and just add the `FilesModel::findMultipleFoldersByFolder()` method in this pull request, I could merge it (if there are not other objections).

christianromeni:
Ok, will do. 
Will be ready soon.

leofeyer:
Merged in contao/core-bundle@3f0687fcda39e2294b9e1113585b6552b679396a.

PR #7944, opened 2015-07-29
Update robots.txt

### LINKED ISSUE #7943: [>=3.2] Update robots.txt
Some still working extensions use the non standard ```/system/modules/*/html/``` folder for assets. Actually Google send some mails that the bot can't crawl to this ressource. We block this with the line ```Disallow: /system/```.

It is useful to add ```Allow: /system/modules/*/html/``` to the robots.txt.

Tastaturberuf:
fix #7943 

Quote:
Some still working extensions use the non standard `/system/modules/*/html/` folder for assets. Actually Google send some mails that the bot can't crawl to this ressource. We block this with the line `Disallow: /system/`.

It is useful to add `Allow: /system/modules/*/html/` to the robots.txt.

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -11,3 +11,4 @@ Disallow: /composer.lock
 Disallow: /README.md
 Allow: /system/cron/cron.txt
 Allow: /system/modules/*/assets/
+Allow: /system/modules/*/html/


discordier:
:+1: from my side. I also still have plenty of extensions which load resources from there.
Most of them are backend only and therefore do not suffer from the robots.txt restrictions but I know that in the wild there are plenty of extensions doing the same.

stefansl:
+1

leofeyer:
Merged in ff304cf2067d7ad346d77e959093ee04ed9370e0.

PR #7973, opened 2015-08-18
Fix phpdoc return type of static find* methods.

dmolineus:
[PhpSpec](http://www.phpspec.net/) within a Contao project stops working after the magic methods got documented. It's caused by the wrong usage of `$this` when documenting static methods. It does not make any sense to use it here as there is no instance of the class.

PHPSpec uses the phpdoc to generate their double classes and creates wrappers for the methods.

I guess it was added because of IDE support. But I really prefer keeping such great tools like phpspec working over fixing IDE issues.

This pull request replaces the `$this` return value by the actual used class.


aschempp:
The problem is that extending the class would not work anymore. You should simply use `static` instead of `$this`. It will look stupid (`@method static static findById($id, $opt=array())`) but is works afaik.

dmolineus:
I can change it. Just documented it like the find* methods with multiple values are documented. They use the class name as well.

aschempp:
I think that's because PhpStorm did not understand the `something|static` statement. Can you find out if PhpSpec does support the `static`?

dmolineus:
The autocompletion actually works. Also with inheritance. But the code inspector of PHPStorm 9 still claims that `static` is an undefined class.
phpstorm

leofeyer:
I am generally ok with the changes.

dmolineus:
> Can you find out if PhpSpec does support the static?

Yes it does.

dmolineus:
I can live with both variants. The static return value or the concrete classes. Both has downsides. It's up to you to decide what you prefer. I would apply the needed changes when. 

I also suggest to add `|null` to the return values as all `find*` methods can return null as result.

leofeyer:
> I also suggest to add `|null` to the return values as all `find*` methods can return null as result.

+1

leofeyer:
I vote for keeping the changes as they are (e.g. `static \CalendarEventsModel` instead of `static static`) and adding `|null` as prosed above.

aschempp:
Fine with me :-)

Toflar:
:+1:

leofeyer:
> I would apply the needed changes when.

@dmolineus Since you have offered to apply the changes, may we ask you to update your pull request accordingly?

dmolineus:
@leofeyer Yes, of course. Would love to see it included in Contao 3.5.3. What's your roadmap? When do I have to finish it then (pretty busy atm)?

leofeyer:
As soon as you can. We'll add it to one of the next 3.5.* releases.

dmolineus:
I added the `|null` return values to the phpdoc.

leofeyer:
Merged in ec63b7d457e8aa66a9e838e8416cabdf4f12ff51. The changes must have taken some time, so thank you very much :)

dmolineus:
Honestly, just 5 minutes with a simple preg_replace script. :)

leofeyer:
+1

PR #7989, opened 2015-08-31
Correctly return the target path if one is set

aschempp:
The current code always returns the original image path if the target path is overridden, which makes the whole functionality unuseable.

According to @Toflar this was already fixed some time ago but apparently re-appeared for unknown reason. I encountered this bug in Contao 4.0.3 so we need to fix both ;-)


discordier:
Original issue was #4166 AFAIR.

leofeyer:
Related: https://github.com/contao/core-bundle/issues/337

aschempp:
They're not related. This is a bug in the code that also exists in Contao 3.5!

leofeyer:
Merged in 73771aa8aaa35748cc315539ad10b815bca47ab3.

PR #8017, opened 2015-09-15
remove $(document).ready()

marcobiedermann:
All functions are called at the bottom of the page, which means the document is already ready.

### DIFF
--- a/system/modules/core/templates/jquery/j_accordion.html5
+++ b/system/modules/core/templates/jquery/j_accordion.html5
@@ -2,22 +2,20 @@
 <script src="<?= TL_ASSETS_URL ?>assets/jquery/ui/<?= $GLOBALS['TL_ASSETS']['JQUERY_UI'] ?>/jquery-ui.min.js"></script>
 <script>
   (function($) {
-    $(document).ready(function() {
-      $(document).accordion({
-        // Put custom options here
-        heightStyle: 'content',
-        header: 'div.toggler',
-        collapsible: true,
-        create: function(event, ui) {
-          ui.header.addClass('active');
-          $('div.toggler').attr('tabindex', 0);
-        },
-        activate: function(event, ui) {
-          ui.newHeader.addClass('active');
-          ui.oldHeader.removeClass('active');
-          $('div.toggler').attr('tabindex', 0);
-        }
-      });
+    $(document).accordion({
+      // Put custom options here
+      heightStyle: 'content',
+      header: 'div.toggler',
+      collapsible: true,
+      create: function(event, ui) {
+        ui.header.addClass('active');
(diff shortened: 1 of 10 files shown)


leofeyer:
This may be right, but since we don't know, why not leave the function call there? Does it noticeably affect the performance?

marcobiedermann:
It does not affect the performance that much, just a bit.
I think this is irrelevant and so less code is cleaner code. It will work perfectly fine without the extra function call.

marcobiedermann:
> So to get the best perceived load time, put your script **at the bottom of the page**. […] And if you're going to do that, then there's **no need to use `ready`**, though of course you could if you liked.

http://stackoverflow.com/questions/6026645/document-readyfunction-vs-script-at-the-bottom-of-page

leofeyer:
I know. But can we be sure that the scripts are always at the bottom of the page? What if someone adds a module via insert tag?

Aybee:
What if something is included with `$GLOBALS['TL_BODY'][]`?
I like this domready wrapper as it also is a reminding marker for that the script needs the DOM to be loaded first.

leofeyer:
@contao/developers /cc

Toflar:
We can safely remove them. That's my opinion :)

aschempp:
I would not remove them. Also because they're a safe wrapper for variables and jquery (like so:)

```js
jQuery(document).ready(function($) {
   // all variables wrapped in function and $ is jQuery
});
```

aschempp:
Which means I would remove the outer function ;-)

leofeyer:
But the outer function is shorter, so in terms of page size optimization, it would make more sense to remove the inner one :)

ausi:
IMO we should still listen to *domready*, because someone could rely on the current behaviour. But there is a shorter version:

Before:

```js
(function($) {
	$(document).ready(function() {
		// code…
	});
})(jQuery);
```

After:

```js
jQuery(function($) {
	// code…
});
```

backbone87:
i think ausi's variant is the way to go. keeps the global clean and you get a save jQuery reference and are safe, if your script isnt at the end of the body.

leofeyer:
Is there a shorter version for MooTools as well?

```js
// Current wrapper
(function($) {
    window.addEvent('domready', function() {
        // …
    });
})(document.id);
```

aschempp:
Well you don't need the outer wrapper if you simply use document.id instead of $ in your code

leofeyer:
I know. But that is actually not the answer to the question :)

backbone87:
```js
window.addEvent("domready", function() {
  var $ = document.id;
});
```

leofeyer:
Changed in contao/core-bundle@ef20adfb6917918fd3769f0552bd852ed5039083.

PR #8021, opened 2015-09-16
Support post arrays for {{post::*}} InsertTag

Toflar:
At the moment I cannot output a simple checkbox wizard form field value on the confirmation page of a form. I know the array is not handled recursively but I don't know if this is needed. However, I think simple arrays should be supported and I even consider this a bug because right now it simply outputs `Array` which is complete nonsense :-)

### DIFF
--- a/system/modules/core/library/Contao/InsertTags.php
+++ b/system/modules/core/library/Contao/InsertTags.php
@@ -1205,6 +1205,29 @@ class InsertTags extends \Controller
 							$arrCache[$strTag] = \System::getReadableSize($arrCache[$strTag]);
 							break;
 
+						case 'flatten':
+							if (!is_array($arrCache[$strTag]))
+							{
+								break;
+							}
+
+							$it = new \RecursiveIteratorIterator(new \RecursiveArrayIterator($arrCache[$strTag]));
+							$result = array();
+
+							foreach ($it as $leafValue)
+							{
+								$keys = array();
+								foreach (range(0, $it->getDepth()) as $depth)
+								{
+									$keys[] = $it->getSubIterator($depth)->key();
+								}
+
+								$result[] = implode('.', $keys) . ': ' . $leafValue;
+							}
+
+							$arrCache[$strTag] = implode(', ', $result);
+							break;
+
 						// HOOK: pass unknown flags to callback functions
 						default:
 							if (isset($GLOBALS['TL_HOOKS']['insertTagFlags']) && is_array($GLOBALS['TL_HOOKS']['insertTagFlags']))


leofeyer:
Shouldn't we add an insert tag flag instead, which serializes the array? I don't mean `serialize()` of course, just a recursive version of `implode(', ', …)`.

aschempp:
We've added the `flatten()` method in Notification Center (https://github.com/terminal42/contao-notification_center/blob/13e5d8b79feb079fa84826701116b22c53b55ade/classes/tl_form.php#L108) which does something similar, it also provides tokens for each option separately. That might only be useful for if/else though.

Toflar:
A "recursiveImplode" flag?

leofeyer:
I'd rather call it `|flatten` :)

Toflar:
Yeah, whatever :-) But this would be rather a feature whereas printing an Array is rather a bug :-)

leofeyer:
It is possible to fix bugs by adding features :)

Toflar:
Yeah, then adding the `flatten` flag sounds reasonable to me.

Toflar:
If you consider it a bug, why is it assigned to 4.1.0 when the bug exists in 3.5 as well? Shall I update my PR?

leofeyer:
Sorry, wrong milestone.

Toflar:
Something like that then?

leofeyer:
Yes, basically. I wonder if we should try to map the nested structures by using punctation?

```php
$a = [
    'group1' => [
        'Jim',
        'Eva',
    ],
    'group2' => [
        'Anna',
        'John',
    ],
];

echo flatten($a);

// group1: Jim, Eva; group2: Anna, John
```

Toflar:
And what if it's nested over 4 levels? :D

leofeyer:
I know.

Toflar:
So? :)

leofeyer:
Well, you need to think of something that could work :)

Maybe we should flatten everything from the 4th level and just add it in brackets?

```php
$a = [
    'group1' => [
        'Jim' => [
            'age' => 36,
        ],
        'Eva' => [
            'age' => 28,
        ],
    ],
    'group2' => [
        'Anna' => [
            'age' => 33,
        ],
        'John' => [
            'age' => 30,
        ],
    ],
];

echo flatten($a);

// group1: Jim (age: 36), Eva (age: 28); group2: Anna (age: 33), John (age: 30)
```

Or we simply skip everything from the 4th level, so the string does not get too long?

Toflar:
Yeah, I could live with that. For other use cases you should probably implement your own inserttag. So I implement your proposal? :)

leofeyer:
Which one? Omitting stuff or flattening stuff as of level 4?

Toflar:
You're example exactly as described in https://github.com/contao/core/pull/8021#issuecomment-145853770

discordier:
Why not adding all keys as dot separated strings and have key value pairs then?

Toflar:
I like that idea too.

discordier:
Would at least be easy to parse then.

Toflar:
Updated my proposal then.

leofeyer:
Merged in cf670e9cb946b01a145a48010407e40fef612a6a. Thanks a lot.

PR #8026, opened 2015-09-21
Don' t ignore the custom template in image hyperlinks (fixes #7964)

### LINKED ISSUE #7964: ContentHyperlink with Image link—Cannot use custom template
Not sure if this is by design, but when trying to use a custom `ce_hyperlink_` template as well as an image link, it always defaults back to the `ce_hyperlink_image` template.

I would think you would want to be able to not just override the default `ce_hyperlink_image` but also be able to specify your own. I am using Contao 3.5.2, but this goes back a while. It all comes down to this [code](https://github.com/contao/core/blob/master/system/modules/core/elements/ContentHyperlink.php#L68), which is not respecting the custom template setting.

Steps to reproduce:

1. Create a Hyperlink element.
2. Select "Create an image link"
3. Select a custom template (i.e. ce_hyperlink_mycustomtemplate)
4. On the frontend, no matter what template you select, ce_hyperlink_image is always used

davidmaack:
Using the custom templates if provided, when generating an image hyperlink template.

### DIFF
--- a/system/modules/core/elements/ContentHyperlink.php
+++ b/system/modules/core/elements/ContentHyperlink.php
@@ -64,8 +64,10 @@ class ContentHyperlink extends \ContentElement
 			}
 			elseif (is_file(TL_ROOT . '/' . $objModel->path))
 			{
+				$strTemplate = ($this->customTpl != '') ? $this->customTpl : 'ce_hyperlink_image';
+
 				/** @var \FrontendTemplate|object $objTemplate */
-				$objTemplate = new \FrontendTemplate('ce_hyperlink_image');
+				$objTemplate = new \FrontendTemplate($strTemplate);
 
 				$this->Template = $objTemplate;
 				$this->Template->setData($this->arrData);


Toflar:
:+1:

leofeyer:
This is wrong, because the custom template would also override the `ce_hyperlink` template. There is no backwards compatible way to fix this, which is why we have changed the templates in Contao 4.

davidmaack:
Why is this wrong? The custom template always overrides the template of the current content element, which is not working when the element decides to switch to the 'ce_hyperlink_image' Template.

fritzmg:
> Why is this wrong? The custom template always overrides the template of the current content element, which is not working when the element decides to switch to the 'ce_hyperlink_image' Template.

Some users might rely on the fact that `ce_hyperlink_image` is hardcoded even though they set a custom template (however unlikely that is). Thus it would break the backwards compatibility.

aschempp:
This has already been changed in Contao 4 if I remember correctly.

leofeyer:
Yes, it has been changed in Contao 4. There is not backwards compatible way to fix this.

PR #8091, opened 2015-11-04
Map fileTree to regular upload widget

katgirl:
It's my second system where you must upload a file at registration


leofeyer:
Fixed in c6cc1fb35869e508c4a059dbbe5179e954073075.

PR #8102, opened 2015-11-10
convertRelativeUrls - don't convert tel: & sms: in href

matzke:
... alternatively we could use a const or else so we are able to put search params in a config

### DIFF
--- a/system/modules/core/library/Contao/Controller.php
+++ b/system/modules/core/library/Contao/Controller.php
@@ -1187,7 +1187,7 @@ abstract class Controller extends \System
 			$strAttribute = $arrUrls[$i+2];
 			$strUrl = $arrUrls[$i+3];
 
-			if (!preg_match('@^(https?://|ftp://|mailto:|#)@i', $strUrl))
+			if (!preg_match('@^(https?://|ftp://|mailto:|tel:|sms:|#)@i', $strUrl))
 			{
 				$strUrl = $strBase . (($strUrl != '/') ? $strUrl : '');
 			}


Toflar:
> . alternatively we could use a const or else so we are able to put search params in a config

I don't think that makes sense. Why should `sms:` ever be converted? :) I think you could extend your list by `skype:` and `sftp`, couldn't you?

leofeyer:
Fixed in ae54893b849db287f81a156d2b63f2706eff7ebf.

discordier:
IMO this is wrong. There are so many schemes more than the ones now implemented.
IMO we should rather convert everything that does not begin with a scheme.
What about:
```php
if (!preg_match('@^([a-z0-9]+:|#)@i', $strUrl))
```

Toflar:
Good idea!

leofeyer:
> There are so many schemes more than the ones now implemented.

For example?

discordier:
http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

leofeyer:
I meant which other protocols are relevant in a web site context?

discordier:
At a quick glance the following make sense (especially when viewing from an intranet point of view):
- ldap
- chrome
- dns
- jabber
- news
- nntp
- ws and wss (when working in javascript context for websockets - must not be rewritten)
- irc
- irc6
- mumble
- smb
- ssh
- teamspeak
- webcal

leofeyer:
I have implemented your suggestion in bda45bcea4e3160a80bf4cfbb60f88dfe945386e.

discordier:
Thanks a lot :+1:, have you (unit-) tested it?

leofeyer:
I have ~~(unit-)~~ tested it.

leofeyer:
But since it was your suggestion, I assume that you have :)

matzke:
Excellent! Great discussion. Good work. Thank you!

leofeyer:
Unfortunately, there is a related problem to this:

https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Idna.php#L127

@discordier How do we know which protocols require `://` and which require only `:`?

leofeyer:
Never mind, I have fixed it myself in 9e7e05d88d98294bd793a3c3ad20a54b86d125ca.

PR #8103, opened 2015-11-11
Update core-uncompressed.js

MoniqueHahnefeld:
I add a new option to openModalSelector() named 'insTagStr'. It's optional. So you can use the pageselector without an inserttag or a custom inserttag. The default inserttag is as it wars 'link_url'. So there won't be conflicts with the new option. I hope you like that and can merge it.

### DIFF
--- a/assets/contao/js/core-uncompressed.js
+++ b/assets/contao/js/core-uncompressed.js
@@ -934,7 +934,7 @@ var Backend =
 				field.value = val.join(',');
 				// #8103 explain returnCallback()
 				if (opt.returnCallback instanceof Function) {
-				    field.value = returnCallback(field.value);
+				    field.value = opt.returnCallback(field.value);
 				} else if (frm.document.location.href.indexOf('contao/page.php') != -1) {
 				    field.value = '{{link_url::' + field.value + '}}';
 				}
--- a/assets/contao/js/core-uncompressed.js
+++ b/assets/contao/js/core-uncompressed.js
@@ -914,7 +914,7 @@ var Backend =
 		M.addButton(Contao.lang.apply, 'btn primary', function() {
 			var frm = window.frames['simple-modal-iframe'],
 				val = [], 
-				inp, field, i, insTagStr, addDefaultInsTag;
+				inp, field, i;
 			if (frm === undefined) {
 				alert('Could not find the SimpleModal frame');
 				return;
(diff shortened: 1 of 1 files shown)


aschempp:
Can you explain the need for this?

MoniqueHahnefeld:
Yes, of course! In my Custom Attribut I want to use the Pagepicker. The Contao Core Pagepicker uses the file page.php and I call the function openModalSelector() with a click on a Button (Also like core). So all is fine, I'll get my id from the selected Page. But this function adds the Insert-Tag to my Value because I use the page.php. But I don't wan't or need this Inset-Tag for my case and I have no chance to clean may value before it will go to my input field. And only of this I don't want to create an own custom page.php file. I can imagine, that other will have this problem too :)

aschempp:
Sounds like a problem we should have had, too. How does Isotope solve this? Maybe @qzminski can give some input, I'm sure he had this problem before :-)

MoniqueHahnefeld:
So there is no other or better solution, or? @aschempp @qzminski 
It makes the code better and adds a nice feature. Why you can not merge this?

qzminski:
Does this code work with regular page pickers now? The ```addDefaultinsTag``` variable is set to true only if the ```opt.insTagStr``` is not empty and it would be empty in the default page picker config.

MoniqueHahnefeld:
@qzminski No if it is not set it is false and not an empty string. that's an difference. I have check this. You can check this too, if you like :)

qzminski:
Well I have just checked and if that option is not set then ```opt.insTagStr``` returns ```undefined``` and thus  ```addDefaultInsTag``` returns true indeed (because ```undefined != ""```). However I would use ```opt.hasOwnProperty('insTagStr')``` check instead as it seems more appropriate to me.

Other than that, I think the PR is okay and can be merged :+1:

MoniqueHahnefeld:
@qzminski ah ok undefined, yes. But this should be so. For this reason that in all other calls of the function openModalSelector() this property is not defined. And in this function the insertTag 'link_url' is setted as default. So if you want to use the function openModalSelector() without getting back any insertTag you have to define the property insTagStr explicit as empty string ''. So for all calls of openModalSelector() in the old regular way (regular page pickers) this function will work.

Or wait if I understand you right, you opt.hasOwnProperty('insTagStr') will give back 'false' instead of 'undefined' :D ?

qzminski:
It's just that I don't like that comparison because in JS the case is as follows:

```js
console.log(undefined ? true : false); // false
console.log("" ? true : false); // false
console.log(undefined == ""); // false (false == false)
```

On second thought however if this part of script is to be improved, then let's improve it even more. I would suggest to add an option for custom return callback which would be even more flexible:

```js
// Usage
returnCallback: function (value) {
    return value;
    // or
    return '{{my_insert_tag::' + value + '}}';
}

// Source code in method
if (opt.returnCallback instanceof Function) {
    field.value = returnCallback(field.value);
} else if (frm.document.location.href.indexOf('contao/page.php') != -1) {
    field.value = '{{link_url::' + field.value + '}}';
}
```

MoniqueHahnefeld:
@qzminski ok, that's nice with the Function :) 
Should I make a new PR with that implementation or will you =)?

qzminski:
Please do as I am currently work overloaded :-(

MoniqueHahnefeld:
Ok, thank you for that! :)  :+1:

MoniqueHahnefeld:
I commited it @aschempp @qzminski 
https://github.com/MoniqueHahnefeld/core-1/commit/6ae793a8af8b91a724e68d48bc59ccb5aa4a45d0

MoniqueHahnefeld:
oh thx!! I was eyeless :D

discordier:
On a sidenote, couldn't you alter the `page.php` and/or Pagepicker now to also use the tag notation? The whole `elseif` could be dropped then.

MoniqueHahnefeld:
@discordier I don't understand o.Ò

discordier:
I am referring to the `} else if (frm.document.location.href.indexOf('contao/page.php') != -1) {` parts.
I meant, you should make `page.php` pass it's own callback aswell and therefore the check here is redundant then. So everything is using the same stuff and there is not any difference for any part of the code.
This way you immediately know that when it works for the core, it works for anyone else as well.

MoniqueHahnefeld:
I thin the solution from @qzminski is very fine. I don't know if the solution which @discordier intented will be backwards compatible. so can we merge it now :) ??? we can optimize in the future still if it is needed =) @aschempp

discordier:
Of course it is backwards compatible... It only will use the same functionality everywhere instead of keeping separate logic branches.

PR #8164, opened 2015-12-11
do not call fclose() with null parameter

mroi:
When a File object is created and closed without ever being used, `resFile` will be `null` on closing. With PHP7, I now get a warning in the log when this happens:

```
PHP Warning: fclose() expects parameter 1 to be resource, null given
```

This patch fixes the issue by duplicating the `is_resource` test from the destructor.

### DIFF
--- a/system/modules/core/library/Contao/File.php
+++ b/system/modules/core/library/Contao/File.php
@@ -534,7 +534,12 @@ class File extends \System
 	 */
 	public function close()
 	{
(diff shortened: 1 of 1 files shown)


discordier:
Nice catch :+1:

leofeyer:
Merged in 646f155d8e544ad0f561f5dec763912026a54e90.

PR #8167, opened 2015-12-16
Must load DCA to make onundo_callback work

aschempp:
Looks like that callback never worked…


leofeyer:
Strange that nobody noticed before …

leofeyer:
Merged in 7798d645549080aa8216a6141734450e4ca1c91c.

discordier:
The DCA could have been loaded before from anywhere else.
So the callback "could" have been working under certain circumstances.

PR #8168, opened 2015-12-16
Always trigger onversion_callback

aschempp:
The `onversion_callback` is currently triggered by the DC driver. This means they are not always triggered when creating a new version, e.g. when toggling a published button.

No idea why the `onrestore_callback` was already triggered by the `Versions` class but `onversion_callback` is not …

### DIFF
--- a/system/modules/calendar/dca/tl_calendar_events.php
+++ b/system/modules/calendar/dca/tl_calendar_events.php
@@ -1059,7 +1059,7 @@ class tl_calendar_events extends Backend
 			$this->redirect('contao/main.php?act=error');
 		}
 
-		$objVersions = new Versions('tl_calendar_events', $intId);
+		$objVersions = new Versions('tl_calendar_events', $intId, $dc);
 		$objVersions->initialize();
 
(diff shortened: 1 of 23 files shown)


Toflar:
Unfortunately, this is a BC break :(

aschempp:
Please explain

Toflar:
Well, all extensions that use `new Versions()` now are not passing `$dc` as third parameter yet. If somebody subscribed to an `onversion_callback` there, they'll suddenly have no `DataContainer` passed with the callback anymore :(

aschempp:
That's correct, but I'm not sure if that will actually be a problem…

Toflar:
Well a missing $dc in the `onversion_callback` seems to be quite heavy, no?

aschempp:
Dunno, in the onrestore_callback the DC is not available, because all useful information (ID and table) are also passed as arguments. The DC contains no additional information …

leofeyer:
@dmolineus Are you using the `onversion_callback` for your workflow extension?

leofeyer:
See #8237.

PR #8181, opened 2016-01-11
silence mkdir warning that file exists

mroi:
The warning of the file created by `mkdir` already existing is no longer silenced by `@` in PHP7.


aschempp:
I guess we should remove the `@` then?

mroi:
Yes, maybe that is not needed any more. I actually have no idea, why the `@` fails to do its job with PHP7. Seems like a potential PHP bug to me.

leofeyer:
Fixed in 215df5b1dab74e2830228126aed22dabefe1e63c.

PR #8183, opened 2016-01-12
Youtube element improvements

aschempp:


system/modules/core/elements/ContentYouTube.php
```diff
@@ -78,7 +78,7 @@ protected function compile()
 
 		$objFile = new \stdClass();
 		$objFile->mime = 'video/x-youtube';
-		$objFile->path = '//www.youtube.com/watch?v=' . $this->youtube;
+		$objFile->path = 'https://www.youtube.com/watch?v=' . $this->youtube;
```
leofeyer: Why this?

aschempp: Because we should always use secure URLs if available.

Toflar: BTW: we did have this discussion already like 4 times :D

leofeyer:
We should generally use the iframe embedding in the future (also see #8217).

aschempp:
I still think this should make it into 3.5 …

jscholtysik:
+1

leofeyer:
Fixed in fd168c3a2f40df650efd41cd4e1f5009e95c90b1.

PR #8193, opened 2016-01-17
Syntax fixes for PHP 7 compatibility

madbonkey:
Hello,

I just tested a Contao installation under PHP 7.0.1. The only thing I had to fix to make it mostly run was change the way a method gets called on a static import in `system/initialize.php`, and it worked mostly fine.
- Contao 3.5.6
- Composer extension
- PHP 7.0.1-nmm1

### DIFF
--- a/system/initialize.php
+++ b/system/initialize.php
@@ -240,7 +240,10 @@ if (isset($GLOBALS['TL_HOOKS']['initializeSystem']) && is_array($GLOBALS['TL_HOO
 {
 	foreach ($GLOBALS['TL_HOOKS']['initializeSystem'] as $callback)
 	{
-		System::importStatic($callback[0])->{$callback[1]}();
+		call_user_func([
+			System::importStatic($callback[0]), 
+			$callback[1]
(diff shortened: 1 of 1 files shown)


NicolasSchneider:
There also seem to be a few namespace issues. I'll start fixing those over the next few days.

aschempp:
What error message did you get about that? I remember that should already be fixed?

leofeyer:
This fix looks wrong to me.

aschempp:
The code certainly works, but I wonder why it would be necessary.

discordier:
Me non neither. PHP 7 should already be supported.
I wonder what the original error message was.

Toflar:
Agreed, I don't think this was the problem either.

nico-schneider:
You can close this, seems to be isolated to my installation of PHP 7. It's a syntax error at `importStatic()->{`, at the curly brace, which it apparently doesn't expect.

nico-schneider:
Also, I can apparently close this.

PR #8215, opened 2016-02-09
Set field and table when generating widget

aschempp:
Without setting the table and field, the widget can never get the correct empty value (see https://github.com/aschempp/contao-core/blob/dcd9187b178a57a8c6e378ca0530e24140a508f6/system/modules/core/modules/ModulePersonalData.php#L287).


discordier:
Anyone knows why they were missing in the first place?

leofeyer:
Merged in fa5fc7710683fcbc385b6642f8319166c47b25ed.

PR #8216, opened 2016-02-09
If a numeric field can be null, the default value is null

aschempp:


discordier:
Seems legit. :+1:

leofeyer:
That's actually not correct. If a field can be `NULL`, it does not necessarily mean that `NULL` is also the default value. Only if the SQL definition says `DEFAULT NULL`, we can safely assume that the default value is `NULL`. Right?

aschempp:
That's correct, but a new handling. Currently we're not checking for the default value at all…

leofeyer:
Merged in 293ab5b55f5800bf030caf834d1cbcd0332868e8. I have also made the null check the first condition to check. Do you remember why you have not?

aschempp:
Probably because the whole thing is hacky :D

PR #8237, opened 2016-02-26
[RFC] Add the oncreate_version and onrestore_version callbacks

leofeyer:
This PR adds the `oncreate_version_callback` and `onrestore_version_callback` as discussed in #8168. You should probably ignore whitespace when looking at the diff :)

@contao/developers Please review.

Diff view with whitespace ignored: https://github.com/contao/core/pull/8237/files?w=1

### DIFF
--- a/system/docs/CHANGELOG.md
+++ b/system/docs/CHANGELOG.md
@@ -4,6 +4,10 @@ Contao Open Source CMS changelog
 Version 3.5.7 (2016-02-XX)
 --------------------------
 
+### New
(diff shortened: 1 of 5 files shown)


aschempp:
Looks good, apart from the fact that I don't like the name (but I have no better idea either…).

discordier:
Same for me and thanks that we are staying bc now. :+1:

leofeyer:
Merged in 81bf4913b8829397104c269545bd92d1c23578a6.

PR #8238, opened 2016-02-26
[RTM] Always fix the domain and language when generating URLs

### LINKED ISSUE #8204: Fehlerhafte URL bei News/Event RSS-Feeds - 3.5.6
Situation:

Events/News haben eine Weiterleitung auf eine Seite/Artikel
Die Sprache wird der URL hinzugefügt

Der RSS Feed enthält bei diesen Events/News eine URL ohne Sprachenkürzel

News.php
https://github.com/contao/core/blob/master/system/modules/news/classes/News.php#L357-L370

Events.php
https://github.com/contao/core/blob/master/system/modules/calendar/classes/Events.php#L406-L419

Lösung für News.php bei interner Seite:
```
			// Link to an internal page
			case 'internal':
				if (($objTarget = $objItem->getRelated('jumpTo')) !== null)
				{
					if (\Config::get('addLanguageToUrl'))
					{
						$obj = \Pagemodel::findWithDetails($objTarget->pid);
						$strForceLang = $obj->language;
					}
					return $strBase . $this->generateFrontendUrl($objTarget->row(), NULL, $strForceLang );
				}
				break;
```

Das Problem ist in der Demo nachstellbar. Contao 3.5.6

leofeyer:
This PR fixed #8204 and all duplicate tickets for good (hopefully).

### DIFF
--- a/system/modules/core/classes/Backend.php
+++ b/system/modules/core/classes/Backend.php
@@ -665,9 +665,14 @@ abstract class Backend extends \Controller
 	 */
 	public static function findSearchablePages($pid=0, $domain='', $blnIsSitemap=false)
 	{
-		$objPages = \PageModel::findPublishedByPid($pid);
+		$time = \Date::floorToMinute();
+		$objDatabase = \Database::getInstance();
 
-		if ($objPages === null)
+		// Get published pages
+		$objPages = $objDatabase->prepare("SELECT * FROM tl_page WHERE pid=? AND (start='' OR start<='$time') AND (stop='' OR stop>'" . ($time + 60) . "') AND published='1' ORDER BY sorting")
+								->execute($pid);
+
+		if ($objPages->numRows < 1)
 		{
 			return array();
 		}
@@ -679,11 +684,12 @@ abstract class Backend extends \Controller
 		}
 
 		$arrPages = array();
-		$time = \Date::floorToMinute();
 
 		// Recursively walk through all subpages
-		foreach ($objPages as $objPage)
+		while ($objPages->next())
 		{
+			$objPage = new \PageModel($objPages);
+
 			if ($objPage->type == 'regular')
 			{
 				// Searchable and not protected
@@ -702,7 +708,10 @@ abstract class Backend extends \Controller
 						$arrPages[] = $feUrl;
 
 						// Get articles with teaser
-						if (($objArticles = \ArticleModel::findPublishedWithTeaserByPid($objPage->id)) !== null)
+						$objArticles = $objDatabase->prepare("SELECT * FROM tl_article WHERE pid=? AND (start='' OR start<='$time') AND (stop='' OR stop>'" . ($time + 60) . "') AND published='1' AND showTeaser='1' ORDER BY sorting")
+												   ->execute($objPages->id);
+
+						if ($objArticles->numRows)
 						{
 							$feUrl = $objPage->getFrontendUrl('/articles/%s');
 
@@ -711,9 +720,9 @@ abstract class Backend extends \Controller
 								$feUrl = $domain . $feUrl;
 							}
 
-							foreach ($objArticles as $objArticle)
+							while ($objArticles->next())
 							{
-								$arrPages[] = sprintf($feUrl, (($objArticle->alias != '' && !\Config::get('disableAlias')) ? $objArticle->alias : $objArticle->id));
+								$arrPages[] = sprintf($feUrl, (($objArticles->alias != '' && !\Config::get('disableAlias')) ? $objArticles->alias : $objArticles->id));
 							}
 						}
 					}
--- a/system/modules/core/classes/Backend.php
+++ b/system/modules/core/classes/Backend.php
@@ -660,20 +660,14 @@ abstract class Backend extends \Controller
 	 * @param integer $pid
 	 * @param string  $domain
 	 * @param boolean $blnIsSitemap
-	 * @param string  $strLanguage
 	 *
 	 * @return array
 	 */
-	public static function findSearchablePages($pid=0, $domain='', $blnIsSitemap=false, $strLanguage='')
+	public static function findSearchablePages($pid=0, $domain='', $blnIsSitemap=false)
 	{
-		$time = \Date::floorToMinute();
-		$objDatabase = \Database::getInstance();
+		$objPages = \PageModel::findPublishedByPid($pid);
 
-		// Get published pages
-		$objPages = $objDatabase->prepare("SELECT * FROM tl_page WHERE pid=? AND (start='' OR start<='$time') AND (stop='' OR stop>'" . ($time + 60) . "') AND published='1' ORDER BY sorting")
-								->execute($pid);
-
-		if ($objPages->numRows < 1)
+		if ($objPages === null)
 		{
 			return array();
 		}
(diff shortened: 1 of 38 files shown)


aschempp:
Why do we need still the `$strForceLang`?

aschempp:
Also, can we add support for `$arrRow` being a `PageModel`?

leofeyer:
Thought about it, too. Maybe not in the bugfix release.

aschempp:
What's the difference? If we always force domain, we always force the language…

leofeyer:
> Why do we need still the `$strForceLang`?

We do not need it, we just keep supporting it for BC. Not sure what people have used it for …

Toflar:
> Not sure what people have used it for …

To force the language, like the parameter says? If you don't have any front end context, how would you generate it correctly? :-)

BTW: I think I have a better approach for your code performance wise which will be fully backwards compatible:

* Check if `tl_page.language` has changed in `save_callback` and if so, update all child records with it.
* Modify the `PageModel` and introduce a `getLanguage()` method:

```php
public function getLanguage()
{
    // Ensure root page language is inherited if not set by save_callback (BC)
    if ('' === $this->language) {
        $this->loadDetails();
    }

    return $this->language;
}
```
* Support passing the `\PageModel` to `\Controller::generateFrontendUrl()` then instead of using `$arrRow['language']` use `\PageModel::getLanguage()`.

Why in 3.5: Well, we have to change the handling because it's buggy at the moment but if we already change it then we should make it right :-)

leofeyer:
> To force the language, like the parameter says? If you don't have any front end context, how would you generate it correctly? :-)

No front end context is required. It will set the language and domain based on `$page->rootLanguage` and `$page->domain`, which are both set in `PageModel::loadDetails()`.

> I think I have a better approach for your code performance wise which will be fully backwards compatible

How is your code performing better? We'd still have to bubble up to the root page to check the domain, wouldn't we?

leofeyer:
Not to forget `$page->trail`, which also required to load the page details.

leofeyer:
The PR is ready to merge now. Please review one last time.

system/modules/core/classes/Backend.php
```diff
@@ -660,20 +660,21 @@ protected function getBackendModule($module)
...
 	 * @return array
 	 */
-	public static function findSearchablePages($pid=0, $domain='', $blnIsSitemap=false, $strLanguage='')
+	public static function findSearchablePages($pid=0, $domain='', $blnIsSitemap=false)
 	{
-		$time = \Date::floorToMinute();
-		$objDatabase = \Database::getInstance();
-
-		// Get published pages
-		$objPages = $objDatabase->prepare("SELECT * FROM tl_page WHERE pid=? AND (start='' OR start<='$time') AND (stop='' OR stop>'" . ($time + 60) . "') AND published='1' ORDER BY sorting")
```
aschempp: If I remember correctly, we intentionally do not use models here because of performance and memory …

system/modules/core/classes/Frontend.php
```diff
@@ -469,7 +469,7 @@ protected function jumpToOrReload($intId, $strParams=null, $strForceLang=null)
 			{
 				if ($intId['id'] != $objPage->id  || $blnForceRedirect)
 				{
-					$this->redirect($this->generateFrontendUrl($intId, $strParams, $strForceLang));
```
aschempp: We must forward the language parameter if it's set, or it would potentially alter behavior.

system/modules/core/library/Contao/Controller.php
```diff
@@ -1075,14 +1075,23 @@ public static function redirect($strLocation, $intStatus=303)
...
 		$strUrl = '';
 
+		if (func_num_args() > 2)
+		{
+			@trigger_error('Using Controller::generateFrontendUrl() with more than two arguments has been deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
+		}
+
+		if (!isset($arrRow['rootId']))
+		{
+			$arrRow = \PageModel::findWithDetails($arrRow['id'])->row();
```
aschempp: I think we should only fetch the domain (and maybe language) here. We don't know if the alias in the array is actually the same as the given page ID…

system/modules/core/models/ArticleModel.php
```diff
@@ -203,6 +203,34 @@ public static function findPublishedByPidAndColumn($intPid, $strColumn, array $a
...
+	 * @param array   $arrOptions An optional options array
+	 *
+	 * @return \Model\Collection|\ArticleModel[]|\ArticleModel|null A collection of models or null if there are no articles in the given column
+	 */
+	public static function findPublishedWithTeaserByPid($intPid, array $arrOptions=array())
+	{
+		$t = static::$strTable;
+		$arrColumns = array("$t.pid=? AND $t.showTeaser=1");
+
+		if (!BE_USER_LOGGED_IN)
```
aschempp: Checking for the backend user is a new and undesired case here, because the searchable pages must not be related to the user being logged in.

system/modules/core/modules/ModuleLogin.php
```diff
@@ -76,7 +76,8 @@ public function generate()
 				// Redirect to the jumpTo page
 				if ($this->jumpTo && ($objTarget = $this->objModel->getRelated('jumpTo')) !== null)
```
aschempp: A general idea. If we would write `($objTarget = $this->objModel->getRelated('jumpTo')) instanceof \PageModel`, it would be more safe, we would not need the `@var` hint and work exactly the same.

system/modules/core/classes/Backend.php
leofeyer: I don't remember anything alike. Has there been a ticket?

system/modules/core/classes/Frontend.php
leofeyer: Why should we? It would only trigger the deprecation warning.

system/modules/core/library/Contao/Controller.php
leofeyer: Isn't this exactly what `findWithDetails()` does?

system/modules/core/classes/Frontend.php
aschempp: it would potentially alter behavior, because it could already be set. We're doing the same in `PageModel::getFrontendUrl()`

system/modules/core/models/ArticleModel.php
leofeyer: This is the default code snippet to make the front end preview work, so I don't think it is "undesired". What is the exact error case?

system/modules/core/modules/ModuleLogin.php
leofeyer: I like the idea. We should generally change this in version 4.2.

leofeyer: I have created a ticket: https://github.com/contao/core-bundle/issues/451

system/modules/core/classes/Frontend.php
leofeyer: I have changed it in 57abb98453bb2673965fc11c3f33eb7c3bd9b3c6.

system/modules/core/classes/Backend.php
Toflar: I remember something like this as well. Cannot find it though but it sure is an issue with a lot of pages. Please do not change this!

system/modules/core/library/Contao/Controller.php
aschempp: Yes, but it will overwrite whatever was in `$arrRow` passed into the method.

system/modules/core/models/ArticleModel.php
aschempp: It is not the same behavior as it previously was. In this implementation, invisible articles would be added to the search index if `BE_USER_LOGGED_IN` is `true` (for whatever reason).

system/modules/core/classes/Backend.php
leofeyer: I have to change it to fix the initial problem.

Toflar: This will fail miserably, I tell you. This is a no go, I'm sorry.

system/modules/core/models/ArticleModel.php
leofeyer: They would only be added if `BE_USER_LOGGED_IN` is `false`, which is never the case in the back end. Also, this is a general model method, which is **not** specifically tailored to the `findSearchablePages()` method!

system/modules/core/classes/Backend.php
leofeyer: You are telling me this based on something you think you remember, which however is nowhere documented?

leofeyer:
I have investigated a bit. Here are some findings of where not to use models:

* https://github.com/contao/core/blob/master/system/modules/core/forms/Form.php#L525-L526
* https://github.com/contao/core/blob/master/system/modules/core/classes/Theme.php#L560-L561

Rebuilding the search index is not one of them. Also, I have debugged rebuilding the search index with and without models:

**With models**

with-models

**Without models**

without-models

Using models requires 10% more memory, however, it also decreases the number of database queries and the number of rows which need to be transferred.

@Toflar What exactly are your concerns?

Toflar:
Try this on an installation with 3000 pages that are split over 12 domains times 4 languages (= 48 root pages). I think it's going to fail :smile:

system/modules/core/classes/Backend.php
aschempp: https://github.com/contao/core/issues/7928#issuecomment-130788699

system/modules/core/models/ArticleModel.php
aschempp: > Also, this is a general model method, which is not specifically tailored to the findSearchablePages() method!

Which is the problem in my opinion. Just saying, you're changing the system behavior in a patch release, that's a very bad idea…

leofeyer:
I found the ticket about what you think to remember: #7928

We have also discussed that we need to use the database class again, but then we found [a bug](https://github.com/contao/core-bundle/commit/faeea6e7ffa37ede2a21f62213fe01047815d00f) in the `Model` class, which turned out to cause the problems.

After fixing the bug, the memory issues were gone. So we can safely merge this PR now.

system/modules/core/models/ArticleModel.php
leofeyer: I am not changing anything, because `BE_USER_LOGGED_IN` **will always be true** in the back end!

aschempp: I hope you mean the opposite? :laughing:

leofeyer: No?

aschempp:
> After fixing the bug, the memory issues were gone.

That's just not true. The last comment was "please test", so we don't know if anything changes. The major difference (in my opinion) is that the models will be kept in the registry indefinitely, whereas the database result will be removed from memory when not used anymore.

system/modules/core/models/ArticleModel.php
aschempp: Well …
1. `BE_USER_LOGGED_IN` **is always false in the backend**
2. Would it be true, the `start`/`stop` filter would not take place, which would be wrong in the `getSearchablePages` context.

leofeyer:
Of course they were gone. We have tested the change extensively, we have just not commented in the ticket. But it was only closed two weeks after finding the bug.

system/modules/core/models/ArticleModel.php
leofeyer: You are right. My bad.

leofeyer: But it does not change the fact that the PR does not change the current behavior in the back end.

leofeyer: If someone executes `Backend::findSearchablePages()` in the front end, it might include unpublished pages now. But you should not have done this anyway (the class is called `Backend` for a reason).

aschempp: That's what I mean. We just don't know. To my gut feeling, we're changing too much, we should just fix the initial problem and not start adding new behavior.

aschempp: Also be aware that your current comparison will result in a wrong result if the constant is not defined (e.g. for a custom entry point).

system/modules/core/library/Contao/Controller.php
leofeyer: I see. Changed in 7bf3c75967f0c8901c8e14e532b46a907bd50f9b.

Toflar:
I tried to do some tests to back my use case and ended up with a regular page model requiring more or less 20kb of RAM. So for 3000 pages it would require about 60 MB RAM which is fine then I think :-) If they are split over 48 root pages it will certainly need quite a lot more memory but it should still be fine. So I guess my concerns are refuted :-) 

However, I still think, inheriting `dns`, `language` and whatever other columns are needed to generate a front end url, is the way to go. You can safely generate a front end url from one single database row without having to load the model and its root page and bubble up all the pages to load the details then. It would greatly improve performance, memory consumption and developer experience :-)

aschempp:
> However, I still think, inheriting dns, language and whatever other columns are needed to generate a front end url, is the way to go. You can safely generate a front end url from one single database row without having to load the model and its root page and bubble up all the pages to load the details then. It would greatly improve performance, memory consumption and developer experience :-)

That's certainly something we should think about, but certainly not in the 3.5.7 patch release ^^

leofeyer:
> However, I still think, inheriting `dns`, `language` and whatever other columns are needed to generate a front end url, is the way to go.

Yes, I'm totally with you on this one. However, that is a different issue.

The problem is that storing the fields in the database is pretty cumbersome and also error-prone, because you have to keep it in sync when the root page changes. There are some solutions for this problem already and backbone87 has implemented one in #6193 (his initial issue was #6830).

leofeyer:
So now that we have settled this and that I have changed all you have suggested, can I merge the PR?

Toflar:
Imho, yes.

leofeyer:
Ah, no, I cannot. I have not solved [this issue](https://github.com/contao/core/pull/8238#discussion_r54392977) brought up by @aschempp :(

system/modules/core/models/ArticleModel.php
leofeyer: I have reverted the changes in a83463a9015027d76706e41a920649bded8b1124.

leofeyer:
> You could have added a flag to $arrOptions to disable the fePreview check?

I have thought about exactly this, too :smile:

I have not asked you though, because I also think it is a new feature. But we should definitely implement it!

leofeyer:
I have created a separate ticket: contao/core-bundle#452

leofeyer:
Merged in c68d813969e0c7c4ae6b1716a71ea6edba00bb8e.

PR #8240, opened 2016-03-01
allow <picture> tag in custom HTML

mroi:
I would like to use the `<picture>` tag in custom HTML. I am not entirely sure, if this is safe to allow by default, but given that `<video>` is already here, I think we should be fine.


aschempp:
:+1:

leofeyer:
Merged in 246a300df7934c075adbd6e9cf1c935747fd73b3.

Issue #7945, opened 2015-07-30
Cannot unset string offsets in Controller.php on line 1478


ausi:
In [Controller.php:1478](https://github.com/contao/core/blob/ac68761904694febb7636efacf34c30575e720a0/system/modules/core/library/Contao/Controller.php#L1478) we try to unset the third array item, but `$size` isn’t always an array and this leads to the fatal error “Cannot unset string offsets”. I think we should change to a `deserialize($arrItem['size'], true)` in [Controller.php:1432](https://github.com/contao/core/blob/ac68761904694febb7636efacf34c30575e720a0/system/modules/core/library/Contao/Controller.php#L1432).

Related issue: #7875

leofeyer:
Fixed in d0c2dae8ef50081e1605e83ad61b11b7a87bb569.

fritzmg:
This fix causes a backwards compatibility break since image aren't resized _at all_ anymore, neither in the frontend nor the backend (see #8197), if `$size` is not an array. Instead of wrapping the size processing block in 
```php
if (is_array($size))
{
	…
}
```
you could simply do
```php
$size = deserialize($arrItem['size']) ?: array(0,0,0);
```
(or what @ausi originally suggested, if that's sufficient too)

ausi:
If you use that, it should be `array(0, 0, 'crop')` or `array(0, 0, '')` IMO.

Or even better:

```php
$size = deserialize($arrItem['size'], true) + array(0, 0, 'crop');
```

which also adds default values if `$arrItem['size']` is an empty or shorter array.

fritzmg:
Sure, sounds good. I chose `array(0,0,0)` because that's what Contao writes into the `size` field by default (in Contao >=3.4.0).

ausi:
In my test Contao wrote `a:3:{i:0;s:0:"";i:1;s:0:"";i:2;s:0:"";}` into the database which is `array('', '', '')`

fritzmg:
Ah yes, I misinterpreted it before, it's `array('','','')'.

PR #8244, opened 2016-03-02
Only treat size as an array, if it's not numeric

bytehead:
As mentioned in [this comment](https://github.com/contao/core/commit/4fdbe3fdf8e53e6954bfc48121ebf994749d98b3#commitcomment-16430718):
Before 3.5.7 it was possible to pass `$arrItem['size']` in `addImageToTemplate` as an `ID` from `tl_image_size`.

This change allows to pass an `ID` and makes it an array for an empty size (see #8197).

### DIFF
--- a/system/modules/core/library/Contao/Controller.php
+++ b/system/modules/core/library/Contao/Controller.php
@@ -1463,12 +1463,18 @@ abstract class Controller extends \System
 		}
 
 		$imgSize = $objFile->imageSize;
-		$size = $arrItem['size'];
+		$size = deserialize($arrItem['size']);
 
-		if (!is_numeric($size))
+		if (is_numeric($size))
 		{
-			$size = deserialize($arrItem['size'], true) + array(0, 0, 'crop');
+			$size = array(0, 0, (int) $size);
 		}
+		elseif (!is_array($size))
+		{
+			$size = array();
+		}
+
+		$size += array(0, 0, 'crop');
 
 		if ($intMaxWidth === null)
 		{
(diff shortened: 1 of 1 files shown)


discordier:
Is it really necessary to deserialize twice?
An serialized array should not pass the is_numeric check and therefore the first call seems superfluous to me.

bytehead:
You're right, it's not necessary. I'll change that.

leofeyer:
@ausi /cc

ausi:
I missed that `$arrItem['size']` could be an integer too, sorry.

IMO this pull request would reintroduce bug #7945. We should make sure that `$size` is always an array. I would change it to:

```php
$size = deserialize($arrItem['size']);

if (is_numeric($size))
{
	$size = array(0, 0, (int) $size);
}
else (!is_array($size))
{
	$size = array();
}

$size += array(0, 0, 'crop');
```

bytehead:
Good point @ausi! I didn't notice that.

aschempp:
Would the max size still be checked if the size is passed as integer to the image method?

ausi:
> Would the max size still be checked if the size is passed as integer to the image method?

I don’t think so.

bytehead:
If there is [no `image_size` available](https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Image.php#L899) it won't.

leofeyer:
Merged in 1d481bd71bb97da8e442d91cb14df47b7c7130e0. Thank you @bytehead and @ausi.

bytehead:
Thank you for the quick merge!

PR #8245, opened 2016-03-02
fix MIME type for SVGs in data: URLs

mroi:
The feature of inlining small images into style sheets by way of data: URLs does not work for me with SVG images. I test with Safari and it does not show the SVG, because the MIME type in the data: URL is set to image/svg instead of image/svg+xml. With the latter, everything works as expected.

So I took the liberty to add a `switch` statement here, in case more special handling is needed later for other file extensions.

### DIFF
--- a/system/modules/core/classes/StyleSheets.php
+++ b/system/modules/core/classes/StyleSheets.php
@@ -2275,16 +2275,17 @@ class StyleSheets extends \Backend
 			$objImage = new \File($strImage, true);
 			$strExtension = $objImage->extension;
 
-			// Fix the jpg mime type
-			if ($strExtension == 'jpg')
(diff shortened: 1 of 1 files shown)


aschempp:
Sounds reasonable but can you fix the coding style?

leofeyer:
Never mind, I'll take care of it :)

mroi:
Great, thanks. I copied the style from another switch statement with similar formatting further up in the same file. But this can of course be changed any way you see fit. Maybe you don’t want a switch here at all, if you don’t expect further need for future customization.

leofeyer:
Fixed in 8e6f3f8f27036b0900012e765a6b1456d5800efb.

Issue #8228, opened 2016-02-17
Navigation Menu with Auto-Start-Level


Hendriks:
What about an extra option for the frontend navigation module to set an **auto-start-level**?

I think a frontend nav module which shows **only child pages of the active parent** is something quite oftenly required, at least for websites with larger / deeper page structures... Shouldn't be too hard to implement this as a **core feature**; I think (but not know ;) Don't you think?

Greets, Hendriks

asaage:
+1 for such a feature  - [like that](https://community.contao.org/de/showthread.php?47199-Navigation-mit-Seiten-von-gleicher-Ebene&p=390988&viewfull=1#post390988)

Toflar:
Do I understand this correctly that it would behave just like if you would check the checkbox `Set a reference page` and then instead of having to choose one, it would simply always take the current page as `Reference page`?

Hendriks:
@Toflar: Yes, this sounds like the kind of pretty simple solution I was imagining... This would completely do the job, I guess... So it could be just another checkbox underneath "set reference page" called "reference to current"...

Toflar:
That's very easy, I'll prepare a Pull Request for Contao 4.2.

Hendriks:
Didn't we have this issue (request) beforehand? And you say it's sooo easy... :) Why not implement it for 3.5 then as well? Shouldn't break existing installations, a checkbox which will be "off" by defaut...

Toflar:
Why do we have to answer this question over and over again? Contao 3.5 is feature complete. Bugfix releases never contain any features and **nothing** will make it into 3.5 anymore except for bugfixes.

Hendriks:
Oh... I wasn't aware / thinking of the current state of 3.5... Sorry...

fritzmg:
On that not, it would be nice if Contao also supports the start- and stop-level when a reference page is set in its default navigation module in some future version.

Toflar:
Actually, it does support the stop-level, not the start level though. Is that intended or is that a bug @leofeyer?

leofeyer:
There is no start level when there is a reference page, because the reference page is the start level.

Toflar:
Which makes no sense :smile: Why should I not start on level 2 based on the reference page? :smile: So to be BC, we would need yet another checkbox :smile:

leofeyer:
And probably more smileys :smile:

Toflar:
Yes :+1: Enjoy weekend, guys!

[...]

PR #8265, opened 2016-03-09
implemented contao/core#8228

fritzmg:


aschempp:
why not simply use the current page if `rootPage` is not set?

asaage:
if rootPage is not set it will take the current website-Root i rely on that in most installations.

fritzmg:
> why not simply use the current page if rootPage is not set?

Sure, but this way it's more clear to the user what will happen, I think. I am fine with it either way ;). Also this way it can be integrated in Contao 4.2 without potentially breaking the backwards compatibility - some users might have `defineRoot` set (by accident maybe), but not `rootPage`.

fritzmg:
Ah, this is actually invalid, it does not implement what was requested in the original issue. In the orginal issue, the user wants the __parent__ page of the active page (`$objPage->pid`) to be the root of the navigation - however, I implemented it with just the active page (`$objPage->id`).

Issue #8254, opened 2016-03-06
E_DEPRECATED wird nicht beachtet


css-umsetzung:
Das errorlog von Contao wächst in wenigen Stunden um mehrere MB an, da Contao nicht meine PHP Einstellungen beachtet, das ich keine deprecated Warnungen loggen oder sehen möchte.

Das Problem tritt nun extrem auf, da Deprecated Meldungen in der generateFrontendUrl eingefügt wurden.

fritzmg:
The `generateFrontendUrl` actually creates `E_USER_DEPRECATED` errors.

Regardless of that, I think the bug is within the error handler here: https://github.com/contao/core/blob/3.5.8/system/helper/functions.php#L43 It checks for
```php
if (ini_get('error_reporting') > 0)
```
However, the PHP documentation states, that you should use
```php
if (error_reporting() > 0)
```
See http://php.net/manual/en/language.operators.errorcontrol.php

>  PHP supports one error control operator: the at sign (@). When prepended to an expression in PHP, any error messages that might be generated by that expression will be ignored.

> If you have set a custom error handler function with set_error_handler() then it will still get called, but this custom error handler can (and should) call error_reporting() which will return 0 when the call that triggered the error was preceded by an @. 

Possibly, with `ini_get('error_reporting')` you only get the server's PHP configuration value, but not the runtime value.

See also https://community.contao.org/de/showthread.php?61539-contao-5

fritzmg:
Hm, can't confirm my suspicion on my local web server though.
```php
var_dump(error_reporting());
var_dump(ini_get('error_reporting'));
```
within the `__error` handler will yield
```
int(0)
string(1) "0"
```
when triggering
```php
@trigger_error('Test.', E_USER_DEPRECATED);
```

@css-umsetzung can you insert
```php
var_dump(error_reporting());
var_dump(ini_get('error_reporting'));
```
[here](https://github.com/contao/core/blob/3.5.8/system/helper/functions.php#L41) (before the if condition) and do the same test and post the results here?

Babelfisch:
Den Fehler kann ich bestätigen. Habe ich in Contao `displayErrors` aktiviert, bekomme ich in der functions.php folgende Werte, wenn `@trigger_error` aufgerufen wurde:

```php
echo ini_get('error_reporting'); // 22519
echo error_reporting(); // 0
```

Mit `error_reporting()` in der Abfrage würden die Meldungen also korrekterweise nicht ausgegeben werden.

[...]

PR #8267, opened 2016-03-11
Correctly check error_reporting to fix issue #8254

discordier:
We have to check the error_reporting() that has been set at runtime and
also validate it against the passed error type. Simply checking for a
value greater than 0 is not enough.

### DIFF
--- a/system/helper/functions.php
+++ b/system/helper/functions.php
@@ -39,8 +39,10 @@ function __error($intType, $strMessage, $strFile, $intLine)
 		E_USER_DEPRECATED   => 'Deprecated notice'
 	);
 
-	// Ignore functions with an error control operator (@function_name)
-	if (ini_get('error_reporting') > 0)
+	// Only log errors that have been configured to get logged.
+	// Functions with an error control operator (@function_name) are also omitted as error_reporting() is 0 in these
+	// cases.
+	if (error_reporting() & $intType)
 	{
 		$e = new Exception();
 


discordier:
We have some open questions though:

1. Shall we add `E_USER_DEPRECATED` to the default omitting?
2. Shall we add a new setting to enable displaying of silenced error messages so developers will see them in Contao and not only when using a custom error handler?

Toflar:
> Shall we add a new setting to enable displaying of silenced error messages so developers will see them in Contao and not only when using a custom error handler?

They already do in Contao 4 and in a much nicer way than we will ever be able to build for Contao 3.5. I think we should just leave it as is. Quality will increase once they start developing their extensions for Contao4 :)

discordier:
I know, but this then renders the deprecation messages in Contao 3.5 useless, as there is no stock way to notice them.
They are hidden all the time, even when you say display errors (due to the silencing operator).
In fact I somehow fail to see the benefit for the average Contao 3 developer here.
Only those applying unit tests and therefore having a different error handler which bails on everything would get a chance to notice them. These developers however are not really the current user base of Contao 3 (IMO) and therefore have yet to learn how to "migrate".

Toflar:
Can we add them to the Debug-Bar without too much effort?

discordier:
Nope, errors are not added to the debug bar at all (and as you know, adding such would imply a new feature).

Toflar:
Adding a new setting too...

discordier:
To put it that way, adding the deprecation errors is also a feature, even a feature that is not usable at all currently.
I for myself have no hard feelings by omitting these messages as I personally do not use the Contao error handler at all in my projects. I just thought it might be useful for the average user to pinpoint porential problems.

You have however not commented on my first point above, shall we add them to the exception list? This is only needed when using `trigger_error` without silencing operator.

leofeyer:
Merged in d068cc2c54918242707d0bd121039b33449449fc

PR #8296, opened 2016-03-31
Eigene Attribute mit an das Template weitergeben

katgirl:
Damit man eigene Werte z.B. aus dem Isotope AttributeWizzard an das Template durchreichen kann.

### DIFF
--- a/system/modules/core/forms/FormCheckBox.php
+++ b/system/modules/core/forms/FormCheckBox.php
@@ -199,16 +199,17 @@ class FormCheckBox extends \Widget
 			}
 			else
 			{
-				$arrOptions[] = array
-				(
-					'type'       => 'option',
-					'name'       => $this->strName . ((count($this->arrOptions) > 1) ? '[]' : ''),
-					'id'         => $this->strId . '_' . $i,
-					'value'      => $arrOption['value'],
-					'checked'    => $this->isChecked($arrOption),
-					'attributes' => $this->getAttributes(),
-					'label'      => $arrOption['label'],
-          'raw'        => $arrOption 
+				$arrOptions[] = array_replace(
+				    $arrOption,
+				    array(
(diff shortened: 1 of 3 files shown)


aschempp:
If we add that (and I'm not sure we should), I would prefer something like this:

```php
$arrOptions[] = array_replace(
    $arrOption,
    array(
        'type'       => 'option',
        'name'       => $this->strName . ((count($this->arrOptions) > 1) ? '[]' : ''),
        'id'         => $this->strId . '_' . $i,
        'checked'    => $this->isChecked($arrOption),
        'attributes' => $this->getAttributes(),
    )
);
```

discordier:
I second the remark by @aschempp, a new sub array feels wrong here.

Toflar:
Same here, I'm with @aschempp and @discordier. However, I do see the need for additional data for the options, so I agree we should do something about this here.

katgirl:
change

discordier:
No it seems like the indention is one step too far, isn't it?

aschempp:
I guess @discordier means the coding style, I had to read it several times before understanding :D

Shouldn't we add the same to the group option?

discordier:
Of course I mean the coding style. Otherwise indention does not make sense. ;)

leofeyer:
@contao/developers Do you want me to merge this (with the CS fixed)?

aschempp:
Well I feel it could be handy in some rare cases, and does not harm for the standard behavior.

Toflar:
I'm 👍  on merging this, yes. Also allows e.g. to add images per option etc. A very valid use case.

discordier:
Also a :+1: from me on this one. Very valid use case.

leofeyer:
Implemented in contao/core-bundle@80dd9c97e892bef55c73caea1de420ac8a0cb8b7.

PR #8302, opened 2016-04-06
[RTM] Fix breadcrumb broken in file picker when using it within tl_page

Toflar:
When you want to add a filepicker field to tl_page, the breadcrumb in the filepicker popup does not work. This is because tl_page does always add the pages breadcrumb which just redirects if `node` is set in the URL even if it is not meant for the page but the file tree. Not sure if my fix is correct though.

/cc @contao/developers

### DIFF
--- a/system/modules/core/classes/Backend.php
+++ b/system/modules/core/classes/Backend.php
@@ -840,16 +840,16 @@ abstract class Backend extends \Controller
 		$objSession = \Session::getInstance();
 
 		// Set a new node
-		if (isset($_GET['node']))
+		if (isset($_GET['pn']))
 		{
 			// Check the path (thanks to Arnaud Buchoux)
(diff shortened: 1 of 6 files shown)


leofeyer:
Your fix is ok, but `if ($_GET['do'] == 'page')` seems simpler to me.

Toflar:
Imho it is still fighting the outcome, not the source of the problem. What do you think about renaming `node` to `fileNode` and `pageNode`?

leofeyer:
I'm fine with that as well.

Toflar:
PR updated.

leofeyer:
What about using `pn`, `fn` and `tn` as keys? Would keep the URL shorter.

Toflar:
Updated.

leofeyer:
Merged in 6827dd86c28f03e9d75aa69b7e5525141209fec4.

PR #8323, opened 2016-04-27
Fixed circular reference paste

aschempp:
Clipboard mode is either "cut" or "copy", check for "paste" results in circular reference issue. The problem does not happen in `tl_page` due to a custom `paste_button_callback`, but can easily be seen if that is removed from `tl_page` DCA.


leofeyer:
Merged in 1fa1cf2f367f268020e5439b4942121ea533b239.

PR #8329, opened 2016-04-29
Fix news feeds in multidomain setups (#8327)

fritzmg:
When using Contao in a multidomain setup, the news feed might contain the wrong links to the items (see #8327). This simple change should fix it completely.

### DIFF
--- a/system/modules/news/classes/News.php
+++ b/system/modules/news/classes/News.php
@@ -355,7 +355,7 @@ class News extends \Frontend
 				if (($objTarget = $objItem->getRelated('jumpTo')) !== null)
 				{
 					/** @var \PageModel $objTarget */
-					return $strBase . $objTarget->getFrontendUrl();
+					return $strBase . $objTarget->getAbsoluteUrl();
 				}
 				break;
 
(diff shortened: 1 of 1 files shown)


fritzmg:
I have updated the pull request. Now news entries with redirect targets to pages or articles will now also have the correct absolute link in the news feed.

leofeyer:
The changes are not correct. The `getLink()` method must only return absolute URLs if `$strBase` is set. Also, I'm pretty sure that the calendar bundle uses similar logic and should therefore be fixed as well, shouldn't it?

fritzmg:
> The changes are not correct. The getLink() method must only return absolute URLs if $strBase is set.

The `getLink` method currently uses:
```php
$strBase . $objTarget->getFrontendUrl();
```
so, this might result in an absolute url, a relative URL, or a nonsense URL, because `\PageModel::getFrontendUrl` either returns an absolute URL or a relative URL, depending on the page's root DNS setting and the current base.

So the resulting URL might result in something like
```
http://www.example1.org/http://www.example2.org/news-entry/foo.html
```
in multidomain setups as shown in  #8327.

> Also, I'm pretty sure that the calendar bundle uses similar logic and should therefore be fixed as well, shouldn't it?

Yes, the structure is different, but as far as I can see the news feed generation for events will also be wrong in multidomain setups, because it also uses
```php
$strBase . $objTarget->getFrontendUrl();
```

leofeyer:
`$strBase` needs to be removed of course and replaced by `$blnAbsoluteUrl`.

fritzmg:
But that would break the backwards compatibility, wouldn't it? Someone might use the `getLink` function with a custom `$strBase`.

leofeyer:
Correct. Then we need a fourth parameter I guess.

fritzmg:
Btw. this is also a case where a `getRelativeUrl` function (contao/core-bundle#458) would come in handy. Because if a `$strBase` is set, the URL generated for the page __must__ be relative, otherwise it would be nonsense again.

i.e.
```php
return $strBase ? $strBase . $objTarget->getRelativeUrl() : $objTarget->getAbsoluteUrl()
```

leofeyer:
Fixed in d16c629bcc7e00a0cd7b4da168cd1d950a108a7a.

DVeris:
url: http**s**://site.com

contao 3.2 in feed ($xml .= '<link>' . specialchars($this->link) . '</link>';)
http**s**://site.com

contao3.5 with this fix display http://site.com

PR #8330, opened 2016-05-01
fixed search results from different domains

fritzmg:
### Problem description

In a multidomain installation, if you happen to have a search module on a page with the refrence page set to a page (or website root) from another domain, the search results will not contain the correct domain (they will be relative to the current domain).
### Proposed fix

In this fix I simply fetch the reference page (if available) and check if its domain differs from the current domain. It should be safe to assume that any search result from that reference page should originate from that domain.
### Other ways to fix
- Since the page ID is available for each search result, the domain check could also be done for each individual search result.
- In [FrontendTemplate.php#L321](https://github.com/contao/core/blob/3.5.12/system/modules/core/classes/FrontendTemplate.php#L321)

``` php
'url' => \Environment::get('request'),
```

could be changed to

``` php
'url' => \Environment::get('uri'),
```

so that the saved URL in `tl_search` includes the scheme and host.

### DIFF
--- a/system/modules/core/modules/ModuleSearch.php
+++ b/system/modules/core/modules/ModuleSearch.php
@@ -252,15 +252,24 @@ class ModuleSearch extends \Module
 				$this->Template->page = $page;
 			}
 
+			// Determine base URL
+			$objReference = \PageModel::findWithDetails($intRootId);
+			$strBase = '';
+
+			if ($objReference->domain && \Environment::get('host') !== $objReference->domain )
+			{
+				$strBase = ($objReference->rootUseSSL ? 'https://' : 'http://') . $objReference->domain . TL_PATH . '/';
+			}
+
 			// Get the results
 			for ($i=($from-1); $i<$to && $i<$count; $i++)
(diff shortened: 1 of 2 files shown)


system/modules/core/modules/ModuleSearch.php
```diff
@@ -252,15 +252,29 @@ protected function compile()
...
+			$strBase = '';
+
+			if ($this->rootPage > 0)
+			{
+				$objReference = \PageModel::findWithDetails($this->rootPage);
+
+				// Check if domain from reference page differs from current domain
+				if ($objReference->domain && \Environment::get('host') !== $objReference->domain )
+				{
+					$strBase = ($objReference->rootUseSSL ? 'https://' : 'http://') . $objReference->domain . TL_PATH . '/';
```
fritzmg: I am not sure if using `TL_PATH` is correct here, need to check that.

fritzmg:
Sorry, messed up my PR branch, I'll clean it up shortly...

Issue #3765, opened 2012-01-05
Multi domain cross linking error


Ruudt:
When I have a multi domain setup I want to include a (sub-)menu of domain2 into an article on domain1. This is possible because I created a navigation module with the reference page somewhere within domain2. However, the links that the menu creates are relative to the base page. This means that the links go to the wrong spot (domain1.com/page.html), where it should obviously be domain2.com/page.html. All of the information would be known to Contao, so I wonder why it should not detect the correct domain and prefix the url whenever needed.

Same thing goes for links inserted into text. For example: {{link_url::20}}. This will generate an url because the page is know, but the base page is left out even though it is on another domain.

For the user the domain is external. But it is 1 Contao setup that manages both domain names, so I'd say it should detect when the domains want to interact with eachother...

fritzmg:
This is still not fixed for news and possibly search, forms and events as well, as adressler mentioned. If the link target of a news entry points to a page that is in a different domain, the generated link will simply not be correct.

fritzmg:
The problem is still present in Contao 3.5.0.

PR #8331, opened 2016-05-02
Fix search results from different domains

fritzmg:
### Problem description

In a multidomain installation, if you happen to have a search module on a page with the refrence page set to a page (or website root) from another domain, the search results will not contain the correct domain (they will be relative to the current domain).
### Proposed fix

In this fix I simply fetch the reference page (if available) and check if its domain differs from the current domain. I think it should be safe to assume that the URL for any search result from that reference page originated from that domain.
### Other ways to fix
- Since the page ID is available for each search result, the domain check could also be done for each individual search result.
- In [FrontendTemplate.php#L321](https://github.com/contao/core/blob/3.5.12/system/modules/core/classes/FrontendTemplate.php#L321)

``` php
'url' => \Environment::get('request'),
```

could be changed to

``` php
'url' => \Environment::get('uri'),
```

so that the saved URL in `tl_search` includes the scheme and host.

---

(Sorry for the mess-up in #8330)

### DIFF
--- a/system/modules/core/modules/ModuleSearch.php
+++ b/system/modules/core/modules/ModuleSearch.php
@@ -252,15 +252,29 @@ class ModuleSearch extends \Module
 				$this->Template->page = $page;
 			}
 
+			// Determine base URL
+			$strBase = '';
+
+			if ($this->rootPage > 0)
+			{
+				$objReference = \PageModel::findWithDetails($this->rootPage);
+
+				// Check if domain from reference page differs from current domain
+				if ($objReference->domain && \Environment::get('host') !== $objReference->domain)
+				{
+					$strBase = ($objReference->rootUseSSL ? 'https://' : 'http://') . $objReference->domain . TL_PATH . '/';
+				}
+			}
+
 			// Get the results
 			for ($i=($from-1); $i<$to && $i<$count; $i++)
 			{
 				/** @var \FrontendTemplate|object $objTemplate */
 				$objTemplate = new \FrontendTemplate($this->searchTpl ?: 'search_default');
 
-				$objTemplate->url = $arrResult[$i]['url'];
+				$objTemplate->url = $strBase . $arrResult[$i]['url'];
 				$objTemplate->link = $arrResult[$i]['title'];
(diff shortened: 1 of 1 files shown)


leofeyer:
>  if you happen to have a search module on a page with the refrence page set to a page (or website root) from another domain, the search results will not contain the correct domain (they will be relative to the current domain).

There is a search form on *domain1.com* and when someone enters a search term, they are being redirected to *domain2.com*? And the result URLs link back to *domain1.com*?

Is there a real use-case for this scenario?

fritzmg:
> There is a search form on domain1.com and when someone enters a search term, they are being redirected to domain2.com? And the result URLs link back to domain1.com?

No, they stay on domain1.com, but also get results from other domains (e.g. domain2.com). Those results will not contain the correct link, since they are relative instead of absolute.

> Is there a real use-case for this scenario?

In my case the customer had a website containing of different sections. Each section had its own domain. However, things like news, events etc. were relevant for all sections and also the requirement was to display the search result from all domains, not just the current domain.

For news and events etc. this has finally been fixed with Contao `3.5.7` (and up), but there are still a few other loose ends, where the multidomain application does not work. This has also been mentioned in past issues (e.g. #3765).

I guess the use case is pretty rare though. Also it might require custom coding anyway, if you want to do it via a single search module, since you are otherwise forced to include several search modules on one page in order to cover multiple domains (since each search module only searches in either the current website root or within the given reference page structure).

leofeyer:
We could solve this by replacing the relative paths like `en/content-elements.html` in `tl_search` with absolute URLs such as `http://localhost/contao-3.5/en/content-elements.html`. I don't think we should use `PageModel::findWithDetails()` here for performance reasons.

@contao/developers /cc

fritzmg:
Indeed, that would be option 2 from "Other ways to fix".

Toflar:
Yes of course, the search index should contain the absolute url.

leofeyer:
See contao/core-bundle#491.

tsarma:
What about making this an option in settings.

PR #8335, opened 2016-05-10
Prevent warnings when invalidating the cache

rolandschuetz:
Easyname uses OPcache, but [restricts the opcache_invalidate() usage](https://www.easyname.at/de/support/hosting/183-warum-erhalte-ich-die-warnung-zend-opcache-api-is-restricted). 

Therefore Easyname Hosting the following warning in thrown:

```
Warning: Zend OPcache API is restricted by "restrict_api" configuration directive in system/modules/rocksolid-custom-elements/src/MadeYourDay/Contao/CustomElements.php on line 1262
```

The customer support write:

```
Das Problem ist das die Funktionen ja existieren (opcache_invalidate), aber es kann leider kein Zugriff auf diese Funktionen stattfinden da der opcache von allen Kunden am Webserver geteilt wird.
```

### DIFF
--- a/system/modules/core/library/Contao/Config.php
+++ b/system/modules/core/library/Contao/Config.php
@@ -265,7 +265,7 @@ class Config
 		// Reset the Zend OPcache
 		if (function_exists('opcache_invalidate'))
 		{
-			opcache_invalidate(TL_ROOT . '/system/config/localconfig.php', true);
+			@opcache_invalidate(TL_ROOT . '/system/config/localconfig.php', true);
 		}
 
 		// Reset the Zend Optimizer+ cache (unfortunately no API to delete just a single file)


ausi:
I commented in madeyourday/contao-rocksolid-custom-elements#66 regarding the same issue:

> I think this is a problem of Easyname. They should either remove the method completely so that `function_exists('opcache_invalidate')` returns false or add the option to disable the warning messages.

leofeyer:
I second @ausi's opinion.

Toflar:
I second @ausi's opinion too.

rolandschuetz:
Yeah, I think too it sucks. Just wasn't sure how else to handle it.
So guess I need to adopt the local Contao core for this hoster :(

ausi:
You could point your hoster to this discussion and ask them if they could implement one of the two proposed solutions.

rolandschuetz:
Yeah, I don't think they will change that soon.
Thanks for the feedback!

Issue #8348, opened 2016-05-27
Registered autoitems are declared as unused get 


MoniqueHahnefeld:
Hi,
I would like to use many autoitems in urls of my extension's modul. I registered my autoitems in the $GLOBALS['TL_AUTO_ITEM'].
This exception does not really make sense than. So maybe can there be added this foreach, which asked if the unused-array values are used as autoitems? =)

THX 

#4277
https://github.com/contao/core/blob/e8b344e4a2aee1bd2c53e166bce7955382232b0b/system/modules/core/classes/FrontendTemplate.php#L115-118

	if ($blnCheckRequest && \Input::hasUnusedGet())
		{
				foreach (\Input::getUnusedGet() as $param) {
					if(!in_array($param, $GLOBALS['TL_AUTO_ITEM'])){
						throw new \UnusedArgumentsException();
					}
				}
		}

leofeyer:
See #8349.

PR #8349, opened 2016-05-27
Update FrontendTemplate.php

MoniqueHahnefeld:
#8348

### DIFF
--- a/system/modules/core/classes/FrontendTemplate.php
+++ b/system/modules/core/classes/FrontendTemplate.php
@@ -114,7 +114,11 @@ class FrontendTemplate extends \Template
 		// Check whether all $_GET parameters have been used (see #4277)
 		if ($blnCheckRequest && \Input::hasUnusedGet())
 		{
-			throw new \UnusedArgumentsException();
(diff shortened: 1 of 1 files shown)


leofeyer:
This does not make much sense to me. What's the initial problem?

MoniqueHahnefeld:
The initial problem is that my registered parameter are in the url and not "unused" and this function throw an error. so I can not process the follow kind of url: xxxx/de/produkte/adapter/adapterfinder/camera/nnn/objectiv/xxx.html
May the problem is in this method:
\Input::hasUnusedGet() 
??
get-parameter to declare as unsed instead of they are used is not really good or ?

aschempp:
I agree with @leofeyer. Why are the parameters in `TL_AUTO_ITEM` if they are in your url?

MoniqueHahnefeld:
ok yes I thought I have to prepare the get-parameter within the $arrFragments into the getPageIdFromUrl() . Now I got it work. It was a little bit tricky to get no conflicts with other extension like events, news and so on. But it's just little bit difficult at all I suppose. Thx @leofeyer @aschempp :)

PR #8363, opened 2016-06-14
[RFC] Build the mapper data before looping through the tables

### LINKED ISSUE #8326: No Imagesize assignment possible when importing a theme
When a theme is imported with a module that uses an Inputfield with the 'imageSize' widget, e.g. Random image, the Inputfield will have no value when a predefined imagesize is selected.

In the Theme class at line [590](https://github.com/contao/core/blob/master/system/modules/core/classes/Theme.php#L590) the imageSize ID should be set to the new one in tl_image_size. But this will never happen because the tl_image_size table will be processed after the tl_module table in the XML file and the array `$arrMapper['tl_image_size']` do not exist at this time. 

It works correctly if I change the export function to write the tl_image_size table first and then the tl_module table into the XML file.
```
	/**
	 * Export a theme
	 *
	 * @param \DataContainer $dc
	 */
	public function exportTheme($dc)
	{
		// Get the theme meta data
		$objTheme = $this->Database->prepare("SELECT * FROM tl_theme WHERE id=?")
								   ->limit(1)
								   ->execute($dc->id);

		if ($objTheme->numRows < 1)
		{
			return;
		}

		// Romanize the name
		$strName = utf8_romanize($objTheme->name);
		$strName = strtolower(str_replace(' ', '_', $strName));
		$strName = preg_replace('/[^A-Za-z0-9._-]/', '', $strName);
		$strName = basename($strName);

		// Create a new XML document
		$xml = new \DOMDocument('1.0', 'UTF-8');
		$xml->formatOutput = true;

		// Root element
		$tables = $xml->createElement('tables');
		$tables = $xml->appendChild($tables);

		// Add the tables
		$this->addTableTlTheme($xml, $tables, $objTheme);
		$this->addTableTlImageSize($xml, $tables, $objTheme); // We need the TLImageSize first
		$this->addTableTlStyleSheet($xml, $tables, $objTheme);
		$this->addTableTlModule($xml, $tables, $objTheme);
		$this->addTableTlLayout($xml, $tables, $objTheme);


```

Or we should provess the tl_image_size table first in the import function to support 'old' .cto files.

leofeyer:
This PR fixes #8326. @ausi @agoat /cc

### DIFF
--- a/system/modules/core/classes/Theme.php
+++ b/system/modules/core/classes/Theme.php
@@ -415,6 +415,36 @@ class Theme extends \Backend
 			$tl_image_size = $this->Database->getNextId('tl_image_size');
 			$tl_image_size_item = $this->Database->getNextId('tl_image_size_item');
 
+			// Build the mapper data (see #8326)
+			for ($i=0; $i<$tables->length; $i++)
+			{
+				$rows = $tables->item($i)->childNodes;
+				$table = $tables->item($i)->getAttribute('name');
+
+				// Skip invalid tables
+				if (!in_array($table, array_keys($arrLocks)))
+				{
+					continue;
+				}
+
+				// Loop through the rows
+				for ($j=0; $j<$rows->length; $j++)
+				{
+					$fields = $rows->item($j)->childNodes;
+
+					// Loop through the fields
+					for ($k=0; $k<$fields->length; $k++)
+					{
+						// Increment the ID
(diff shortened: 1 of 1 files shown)


ausi:
Looks good to me. Tested it and it worked as expected.

leofeyer:
Merged in afb01b851eeb06961cec079174e159105465f693.

PR #8364, opened 2016-06-14
[RFC] Skip empty reader URLs when finding searchable pages

leofeyer:
This PR implements a possible solution for #8361. @contao/developers What do you think?

### DIFF
--- a/system/modules/calendar/classes/Calendar.php
+++ b/system/modules/calendar/classes/Calendar.php
@@ -358,10 +358,18 @@ class Calendar extends \Frontend
 						$arrPages[] = sprintf($strUrl, (($objEvents->alias != '' && !\Config::get('disableAlias')) ? $objEvents->alias : $objEvents->id));
 					}
 				}
+
+				$strReaderUrl = str_replace('/%s', '', $strUrl);
+
+				// Remove the empty event reader URL (see #8361)
+				if (($i = array_search($strReaderUrl, $arrPages)) !== false)
+				{
+					unset($arrPages[$i]);
+				}
 			}
 		}
 
-		return $arrPages;
+		return array_values($arrPages);
 	}
 
 
--- a/system/modules/faq/modules/ModuleFaq.php
+++ b/system/modules/faq/modules/ModuleFaq.php
@@ -99,9 +99,17 @@ class ModuleFaq extends \Frontend
 						$arrPages[] = sprintf($strUrl, (($objItems->alias != '' && !\Config::get('disableAlias')) ? $objItems->alias : $objItems->id));
 					}
 				}
+
+				$strReaderUrl = str_replace('/%s', '', $strUrl);
+
+				// Remove the empty FAQ reader URL (see #8361)
+				if (($i = array_search($strReaderUrl, $arrPages)) !== false)
+				{
+					unset($arrPages[$i]);
+				}
 			}
 		}
 
-		return $arrPages;
+		return array_values($arrPages);
 	}
 }
--- a/system/modules/news/classes/News.php
+++ b/system/modules/news/classes/News.php
@@ -325,10 +325,18 @@ class News extends \Frontend
 						$arrPages[] = $this->getLink($objArticle, $strUrl);
 					}
 				}
+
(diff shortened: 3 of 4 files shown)


frontendschlampe:
What will happen, if I navigate to empty reader URL in frontend? Normally there should be a 404 error?! And what is with the sitemap module?

leofeyer:
> What will happen, if I navigate to empty reader URL in frontend?

Right now, no error is triggered. But changing this requires a separate ticket, because this one is about the empty reader URLs being added to the sitemap.xml file.

> And what is with the sitemap module?

The sitemap does not include links to events or news and is therefore not affected by this change.

frontendschlampe:
> Right now, no error is triggered. But changing this requires a separate ticket, because this one is about the empty reader URLs being added to the sitemap.xml file.

ok ... I will do so

> The sitemap does not include links to events or news and is therefore not affected by this change.

ok ... but what's with the empty detail page? This also should not bei listed in the sitemap module.

leofeyer:
> What will happen, if I navigate to empty reader URL in frontend? Normally there should be a 404 error?! 

By the way, I am not sure about this. Right now the module is simply not rendered if there is no news or event ID. People might rely on this behavior.

discordier:
A detail page does not necessarily only contain the detail module.
It might also contain other content, which will then not be reachable anymore.
Therefore it is IMO way too hard to remove the whole page from the searchable pages then.

In the end, I am one of said people that rely on this behaviour in at least some installations.

I mean, I seem to remember that it is possible to have the list module and the reader module on the same page.
If this gets merged, the list page won't get indexed anymore (the reason to do index a list page is questionable, but it serves as an example here of what seems to be possible currently and will break if this get's changed).

asaage:
i agree with @discordier 
i would not want the readerpage to be skiped from sitemap and i would not like to get 404 on an empty readerpage either #8367

frontendschlampe:
> I mean, I seem to remember that it is possible to have the list module and the reader module on the same page.

I think, this is not possible for every function (news, calender etc.).

> Therefore it is IMO way too hard to remove the whole page from the searchable pages then.

Why not? All the information of the empty reader page will be in the other detail pages. It will be negative if there will be some more important content then the details of the entry because there will be many detail pages with the same information. You know what I mean? The Information are not gone, they are just on another page!

asaage:
> The Information are not gone, they are just on another page!

In case you have news or calender-items to display you are right.

Putting the list and the reader module on the same page is something what i do oftentimes with news faq's and events

frontendschlampe:
> Putting the list and the reader module on the same page is something what i do oftentimes with news faq's and events

But this is no problem for the sitemap.xml because the information of list is not gone!

fritzmg:
> A detail page does not necessarily only contain the detail module.
> It might also contain other content, which will then not be reachable anymore.
> Therefore it is IMO way too hard to remove the whole page from the searchable pages then.
> 
> In the end, I am one of said people that rely on this behaviour in at least some installations.
> 
> I mean, I seem to remember that it is possible to have the list module and the reader module on the same page.
> If this gets merged, the list page won't get indexed anymore (the reason to do index a list page is questionable, but it serves as an example here of what seems to be possible currently and will break if this get's changed).

Indeed, that's why I originally posted it as a feature request: #contao/news-bundle#5

I don't really see it as a bug, since this whole thing probably should have customizable options. i.e. whether or not the empty page should be removed from the `sitemap.xml` and whether or not the reader module should redirect to a news, when no `auto_item` parameter is given (or give a 404 page).

fritzmg:
So, since the original feature request (contao/news-bundle#5), the resulting contao/core ticket (#8361) and this pull request are closed now, what is the state on this?

leofeyer:
The PR was closed automatically, because you did not target it towards an infinite branch.

leofeyer:
I have re-opened #8361.

discordier:
> The PR was closed automatically, because you did not target it towards an infinite branch.

@leofeyer it was your PR :smile_cat:

leofeyer:
I know. I was talking to myself 😄

PR #8365, opened 2016-06-14
Set default to NULL on copy if nullIfEmpty is set

aschempp:
I have a date field in my DCA which can be NULL and `doNotCopy` enabled. If the record is duplicated, the value currently results in 01.01.1970.


leofeyer:
Shouldn't we use `Widget::getEmptyValue()` here?

aschempp:
I thought about that too, but that means an integer field would get '0' as default, which is probably not what you want?

leofeyer:
0 seems correct for an INT field IMHO.

aschempp:
But its very likely not what you want in a field, because you want to enter a new value.

What happens if a new record is created? Are we using 0 or empty?

leofeyer:
We are using `Widget::getEmptyValue()` 😄

leofeyer:
Fixed in aff8643c9651ac6061c8e0c414712d10b4d76dfa.

PR #8381, opened 2016-06-24
Feature: "previous"

marcobiedermann:
I suggest renaming all "prev" links to "previous" to make it more consistent and easier to develop themes.

### DIFF
--- a/system/modules/core/templates/modules/mod_booknav.html5
+++ b/system/modules/core/templates/modules/mod_booknav.html5
@@ -4,9 +4,9 @@
 
   <ul>
     <?php if ($this->hasPrev): ?>
-      <li class="prev"><a href="<?= $this->prevHref ?: './' ?>" title="<?= $this->prevPageTitle ?>">&lt; <?= $this->prevLink ?></a></li>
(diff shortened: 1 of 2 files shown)


fritzmg:
I think this is already normalized in Contao 4.

marcobiedermann:
@fritzmg Yes, I know. But it would be great if this would land in Contao 3.x

fritzmg:
That's not possible, because that would break backwards compatibility. If this was integrated and someone updates their Contao installation, their "previous" buttons might not be styled anymore all of a sudden.

marcobiedermann:
@fritzmg Sure. But you could release Contao 3.6 and let people know

aschempp:
There will not be a new Contao 3.x version (see [release plan](https://contao.org/en/release-plan.html))

fritzmg:
> @fritzmg Sure. But you could release Contao 3.6 and let people know

Even if there was a Contao 3.6, it cannot be integrated into Contao 3.6. Minor versions cannot break the backwards compatibility. Things like that can only be changed in major versions.

Issue #8396, opened 2016-07-08
Changing Accordion to work with everything, not only div.toggler and div.toggler


MacKP:
It would be nice if the accordion is more flexible. At the Moment it works only with div Elements.
I will make two PRs for that (jquery and mootools) for the .html5 Templates.

regards

leofeyer:
You could have changed this in one PR, too. 😄

PR #8397, opened 2016-07-08
Changing to work without div

MacKP:
See this ticket: https://github.com/contao/core/issues/8396


leofeyer:
We should also check the MooTools templates.

leofeyer:
Changed in contao/core-bundle@f74d7789885d97cea5d382516c34fc73f10f18bc.

PR #8398, opened 2016-07-08
Changing to work without the div

MacKP:
See Ticket: https://github.com/contao/core/issues/8396

### DIFF
--- a/system/modules/core/templates/mootools/moo_accordion.html5
+++ b/system/modules/core/templates/mootools/moo_accordion.html5
@@ -2,7 +2,7 @@
 <script>
   (function() {
     window.addEvent('domready', function() {
-      new Fx.Accordion($$('div.toggler'), $$('div.accordion'), {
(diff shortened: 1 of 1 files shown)


leofeyer:
I think we are going to have a problem if we do not also adjust [line 11](https://github.com/contao/core/blob/master/system/modules/core/templates/mootools/moo_accordion.html5#L11), which expects the markup to contain DIVs.

@aschempp /cc

aschempp:
no idea what that actually does… do you?

leofeyer:
It selects the DIV after the toggler and fades it in. Maybe `getNext('.accordion')` works?

leofeyer:
> Maybe `getNext('.accordion')` works?

It does actually. 😄

leofeyer:
Changed in contao/core-bundle@f74d7789885d97cea5d382516c34fc73f10f18bc.

PR #8425, opened 2016-08-02
[RTM] Harden authentication against session fixation

discordier:
This hardens the authentication against session fixation attacks.
Thanks to Rico Kaltofen for pointing it out.

### DIFF
--- a/system/modules/core/library/Contao/User.php
+++ b/system/modules/core/library/Contao/User.php
@@ -553,6 +553,9 @@ abstract class User extends \System
 	{
 		$time = time();
 
+		// Regenerate a new session id to harden against session fixation attacks.
+		session_regenerate_id();
+
 		// Generate the cookie hash
 		$this->strHash = sha1(session_id() . (!\Config::get('disableIpCheck') ? $this->strIp : '') . $this->strCookie);
 


leofeyer:
@contao/developers Haven't we already agreed that this fix is wrong and likely to have unforeseen side effects?

discordier:
No, we did not.
Definately not in 3.5, as this simply ensures the session id get's changed (without loosing the values of the session itself).
So there is no side effect possible unless someone else is using `session_id()` in their code prior authentication, keeps that value and does it use after authentication where the session id has been changed then.

Toflar:
I think it should go into 3.5.

leofeyer:
> So there is no side effect possible unless someone else is using `session_id()` in their code prior authentication, keeps that value and does it use after authentication where the session id has been changed then.

Or if someone manually adds a row to `tl_session` e.g. based on a single sign on extension or remember me implementation. I have briefly discussed possible side effects with @aschempp in Mumble yesterday and we cannot assume that there will not be any.

discordier:
>Or if someone manually adds a row to tl_session e.g. based on a single sign on extension or remember me implementation.

In theses cases, the session is already present and authenticated (manual adding the session to the table). So no problem here.
For remember me tokens, why should the code rely upon the session ID(!) to remain the same aftert calling `generateSession`? It does not make sense at all. The session ID is local to the application.

I would be delighted if you could elaborate on your brief discussion here because I fail to see any side effects that would not harm security already.

system/modules/core/library/Contao/User.php
```diff
@@ -553,6 +553,9 @@ protected function generateSession()
 	{
 		$time = time();
 
+		// Regenerate a new session id to harden against session fixation attacks.
```
leofeyer: CS: no full stop at the end.

leofeyer:
Merged in 9ae1234333610687331753bd78d6bea83674a96c.

Issue #8419, opened 2016-07-26
Autofocus on cssId in tl_content edit view


bezin:
Hey Guys, 

in tl_content edit view (with tl_article as the parent), the cssId field gets the autofocus attribute, if the expert settings panel is open. If you want to edit a content element, you sometimes have to scroll all the way up.

You may reproduce this in the online demo by opening a random article and edit any content element inside. If the expert settings panel is open, cssId will be autofocused.

I tested this with Chrome 51 on Windows 10. 

Best
benzin

aschempp:
Is this reproduceable in the online demo? The edit form should autofocus the first input field. Maybe you set eval => autofocus ?

bezin:
As I wrote, it is reproducable in the online demo, but I'll be more specific :)
 
1. Go to the "Home" article on the frontpage.
2. Edit the first content element "About this demo"
> If the expert settings panel is closed, the headline field will be focused. 
> If the expert settings panel is open, the cssId field will be focused. 

Do prove, do the following: 
3. Open the expert settings panel. Just toggle it. 
4. Go back.
5. Edit the same content element again.

The cssId input will be focused – but the headline input should have been focused according to the rule you mentioned.

fritzmg:
I cannot reproduce the behaviour, at least not under Windows 7 x64 with FireFox 32bit 47.0.1. __But__ the actual HTML output does indeed assign the `autofocus` attribute to the `ctrl_cssID_0` input field in the online demo:
```html
<input type="text" name="cssID[]" id="ctrl_cssID_0" class="tl_text_2" value="" autofocus onfocus="Backend.getScrollOffset()">
```

bezin:
I did some browser tests (all on Windows 10 x64, if not stated otherwise): 
The issue occurs as described in Chrome 51, IE 11, Edge, Opera 38.
It works as expected in Firefox 47, as well as in Safari 8 on MacOSX 10.10.

fritzmg:
The problem happens here: https://github.com/contao/core/blob/3.5.15/system/modules/core/drivers/DC_Table.php#L1917
```php
// Autofocus the first field
if ($blnIsFirst && $GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['inputType'] == 'text')
{
	echo 'foo';
	$GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['eval']['autofocus'] = 'autofocus';
	$blnIsFirst = false;
}
```
[...]

PR #8427, opened 2016-08-03
[RTM] Hotfix/autofocus

Toflar:
See #8419.

Screencast GIF (very slow for some reason but I'm too lazy to investigate):
example


leofeyer:
What if the first focusable field is a TinyMCE textarea?

Toflar:
> What if the first focusable field is a TinyMCE textarea?

Implemented support in eba721d79b79e8d3dd690bea5a8eb24ca89f5631. Pretty cool, hm? 😎

leofeyer:
Does this work if `TinyMCE.init()` is called before `Backend.autoFocusFirstInputField()`?

Toflar:
Why should this ever happen?

discordier:
I have an issue with this.
This will prevent me from scrolling though the page with the keyboard arrow keys without prior clicking outside of the autofocus input field or tabbing to the first non input element.
As soon as the cursor is within the field, the browser auto completion will get triggered by the arrow keys  instead of page scrolling.
Maybe a non issue for Mac Users as these are used to scrolling with the pad but definately an issue on other platforms where people use the keyboard where possible.

Toflar:
How is this related to this at all? We have had autofocus ever since so you had that issue ever since too?

discordier:
I do not know about you but in my (stock) 3.5.15, the search field is not autofocussed and hence I can happily scroll up and down through the list views. I wonder why this should be different for me than for you.

In your screen cast, the search field is autofocus'ed and therefore this will not be possible anymore.

fritzmg:
The autofocus attribute doesn't seem to work in all browsers, maybe that's why you didn't have an issue yet.

Toflar:
You weren't talking about the list view. You were talking about scrolling in general.
I see. I actually like the feature but I can understand your concerns. The question is whether we want to skip the filter panel or restrict the focussing to the edit views only (which will cause e.g. the maintenance mode screen not to focus anything at all by default).

discordier:
Maybe you misunderstood what I am referring to.
You mean the auto focus in the EDIT view, that is present and will still be present.
This is not that much of an issue to me as when editing one will most likely want to use the input fields.
I am only talking about the LIST view, where this does not bring any benefit to me but imposes above explained issue.
Scrolling through the article or page list is not possible anymore.

discordier:
Skipping the panels would be very good IMO.

Concerning the maintenance mode... this will then select the "live update ID" input always which is the only input on that screen that will get modified only when performing an upgrade and therefore does not make any sense to have it always auto focus'ed.

Toflar:
This was obviously just a general example...you can take anything you like. The question is whether to skip explicitly the filter panel or limit the focussing feature to the known edit masks (edit, editAll etc.) only.

discordier:
I would limit it to the edit masks.

aschempp:
I would only apply it to the edit mask too.

leofeyer:
Me too.

Toflar:
Removed other autofocus related stuff in bd8d338c070cac131fbcb3b47720da699c311d6d (this was very likely interfering with the autofocus attribute before, so I don't know how this could have ever worked properly at all) and implemented the restriction on all edit masks only in c407b53fccf560d9fcde7da0793a6b12681c307c.

leofeyer:
Again, what happens if `TinyMCE.init()` is called before `Backend.autoFocusFirstInputField()`?

Toflar:
> Again, what happens if TinyMCE.init() is called before Backend.autoFocusFirstInputField()?

Again: Why should this ever happen?

leofeyer:
E.g. if someone moves the scripts down to the end of the page (people are using hooks to accomplish this) or someone adds the `async` attribute.

aschempp:
> E.g. if someone moves the scripts down to the end of the page (people are using hooks to accomplish this) or someone adds the async attribute.

In the backend? 😧

Toflar:
I don't see what you want to tell me tho be honest. The `Backend.autoFocusFirstInputField()` is triggered on `domready` just like all the other stuff like `Backend.makeWizardsSortable()` and  	`Backend.enableToggleSelect()`. If the DOM is ready, the `<textarea>` is gonig to get the focus. That's it. If then `tinyMCE.init()` is executed, it notices the focus on the target element and focuses itself. Your example would require to execute `tinyMCE.init()` before the DOM is ready which seems strange to me.
But I can add the feature so it works in both cases, if you feel better with it :)

leofeyer:
> is triggered on domready just like all the other stuff

You are right, I did not think of this. Then the PR is `RTM` to me.

leofeyer:
Merged in d34662322e4f5d226990279472a7960b88601b87.

PR #8430, opened 2016-08-05
Fixed UUID not present in file upload array when existing file is bei…

Toflar:
…ng replaced.


Toflar:
Ah, duplicate of https://github.com/contao/core/pull/8421. Hotfixes should be published more often... :)

PR #8435, opened 2016-08-11
Allow only those roots that are allowed in root nodes in page selector

qzminski:
The problem is that the pages outside the `rootNodes` scope of a given field could still be accessed by its page picker if they are set in `$GLOBALS['TL_DCA']['tl_page']['list']['sorting']['root']`, e.g. because of breadcrumb.

### DIFF
--- a/system/modules/core/widgets/PageSelector.php
+++ b/system/modules/core/widgets/PageSelector.php
@@ -170,7 +170,7 @@ class PageSelector extends \Widget
 					);
 
 					if (count($root) === 0) {
-						$root = $this->rootNodes;
+						$root = $this->User->hasAccess($this->rootNodes, 'pagemounts') ? $this->rootNodes : [];
 
 						// Hide the breadcrumb
 						$GLOBALS['TL_DCA']['tl_page']['list']['sorting']['breadcrumb'] = '';
--- a/system/modules/core/widgets/PageSelector.php
+++ b/system/modules/core/widgets/PageSelector.php
@@ -159,7 +159,25 @@ class PageSelector extends \Widget
 			// Root nodes (breadcrumb menu)
 			if (!empty($GLOBALS['TL_DCA']['tl_page']['list']['sorting']['root']))
 			{
-				$nodes = $this->eliminateNestedPages($GLOBALS['TL_DCA']['tl_page']['list']['sorting']['root']);
+				$root = $GLOBALS['TL_DCA']['tl_page']['list']['sorting']['root'];
+
+				// Allow only those roots that are allowed in root nodes
+				if (is_array($this->rootNodes))
+				{
+					$root = array_intersect(
+						array_merge($this->rootNodes, $this->Database->getChildRecords($this->rootNodes, 'tl_page')),
+						$root
+					);
+
+					if (count($root) === 0) {
+						$root = $this->rootNodes;
+
+						// Hide the breadcrumb
+						$GLOBALS['TL_DCA']['tl_page']['list']['sorting']['breadcrumb'] = '';
+					}
+				}
+
+				$nodes = $this->eliminateNestedPages($root);
 
 				foreach ($nodes as $node)
 				{


aschempp:
If I understand your PR correctly, all `rootNodes` would be shown if no page is allowed in both arrays? Isn't that a security issue (because DCA `root` is also used for access limitation)?

qzminski:
Good point Andy, just added the permission check.

aschempp:
Unfortunately, the root node array can be modified by any extension that adds further permissions checks we don't know about.…

qzminski:
If extension does affect user pagemount permissions, then it should also alter the ```$this->User->pagemounts``` as it is used in other parts of Contao, too.

aschempp:
> If extension does affect user pagemount permissions, then it should also alter the $this->User->pagemounts as it is used in other parts of Contao, too.

I don't think that would be appropriate. There can be dynamic reasons to do that, imagine for example a backend module that only shows the backend tree of the currently used domain to prevent new users from known issues.

qzminski:
What do you propose then? Yet another hook?

aschempp:
No. First of all we should figure out why `root` is set at all, because apparently there are two different session filters for the regular page view and the picker popup.

qzminski:
Are you sure? In this particular case the root is set by breadcrumb generation but not in the picker itself:

https://github.com/qzminski/core/blob/782f8964a05c09a00283518264edfe77ca9efa6f/system/modules/core/widgets/PageSelector.php#L153-L157

but in the ```dca/tl_page```:

https://github.com/qzminski/core/blob/782f8964a05c09a00283518264edfe77ca9efa6f/system/modules/core/dca/tl_page.php#L28

aschempp:
Aaah, the `$do != 'page'` this is the missing link @Toflar and me didn't see. In this case, the correct solution should be to not set the breadcrumb in the popup, but `array_intersect` both arrays, and possibly not show any page if necessary. What do you think?

qzminski:
Then in the popup you wouldn't be able to use the breadcrumb which we have certainly used when testing the issue. I could live with that but that's limiting the existing pretty useful feature, especially when your website root consists of many many pages.

Besides, the method that generates breadcrumb also seems to ignore the ```root``` setting and completely override it. That means it does not support what you want either.

https://github.com/qzminski/core/blob/782f8964a05c09a00283518264edfe77ca9efa6f/system/modules/core/classes/Backend.php#L930

aschempp:
That's the point. There's a problem here. First of all, the picker should have a different session ID. It actually does, for everything that is not `$do = 'page'`. Secondly, the session ID should probably be differente for each field, don't you think?

qzminski:
Given that we would have took off that ```if ($do != 'page')``` statement and disable default breadcrumb generation in ```tl_page``` for popups, we would have to do the allowed IDs check before generating the breadcrumb then (before the ```root``` is affected). Of course that would make sense only if the session ID is unique to the field and the table (and maybe even to the record ID).

aschempp:
Idk, other opinions @contao/developers ?

Toflar:
Why would the session be unique to the field? I think it makes sense to work on the same page tree in general, it should just make sure that if `rootNodes` are set, they are considered.

aschempp:
> it should just make sure that if rootNodes are set, they are considered.

Does that mean it would ignore `root` filter if `rootNodes` is set?

Toflar:
Only if they are not compatible. If root is allowed in rootNodes then why not display the breadcrumb?

aschempp:
Problem is that "root" does not necessarily mean breadcrumb, it could well be access permissions too. But imagine this in the case of ChangeLanguage, if the user does not have access to english but want's to select an english page for the german page tree, what should happen?

Toflar:
If the user is not allowed to see certain pages, it's irrelevant what `rootNodes` allows. The page picker must only show the intersection of `root` and `rootNodes`.

aschempp:
But that's the problem. He's allowed to see them, they're just limited by his node filter.

@qzminski what happens if we `array_intersect` to an empty value? Can he still see the breadcrumb and maybe remove it?

qzminski:
If one of the ```array_intersect``` parameters is an empty array then you will get back an empty array. You can already see in the code that if the ```$root``` is empty then breadcrumb is hidden.

aschempp:
> if the $root is empty then breadcrumb is hidden

Maybe that's wrong and should be changed then?

qzminski:
Huh? How would you like to change that?

Toflar:
We need this in 3.5.16, what's missing?

system/modules/core/widgets/PageSelector.php
```diff
@@ -159,7 +159,25 @@ public function generate()
...
+				// Allow only those roots that are allowed in root nodes
+				if (is_array($this->rootNodes))
+				{
+					$root = array_intersect(
+						array_merge($this->rootNodes, $this->Database->getChildRecords($this->rootNodes, 'tl_page')),
+						$root
+					);
+
+					if (count($root) === 0) {
+						$root = $this->User->hasAccess($this->rootNodes, 'pagemounts') ? $this->rootNodes : [];
```
leofeyer: This could be simplified as `$root = $this->rootNodes;`.

leofeyer:
After merging, we should also check if the same happens for the file picker.

leofeyer:
Merged in 05d61aee98699adbca6c6d07e111bfd456c0a810.

leofeyer:
Adjusted the file picker in 3329e86c1aceaafb0751f9132afc9e2e5b2c0123.

PR #8438, opened 2016-08-12
[RFC] Fixed simple tokens

Toflar:
Fixed simple tokens parsing ignoring certain characters for now reason and without a trace in the logs.

This is now logged as follows:

bildschirmfoto 2016-08-12 um 12 05 50

Moreover, any character is now allowed for tokens. You can use `-` without any problem. Even using `'` works even though this is very uncommon :-)

Obviously you cannot use any of these characters because they are used for comparison: `=!<>`.

### DIFF
--- a/system/modules/core/library/Contao/StringUtil.php
+++ b/system/modules/core/library/Contao/StringUtil.php
@@ -491,13 +491,22 @@ class StringUtil
 		// Replace the tags
 		foreach ($arrTags as $strTag)
 		{
-			if (strncmp($strTag, '{if', 3) === 0)
+			if (strncmp($strTag, '{if', 3) === 0 || strncmp($strTag, '{elseif', 7) === 0)
 			{
-				$strReturn .= preg_replace('/\{if ([A-Za-z0-9_]+)([=!<>]+)([^;$\(\)\[\]\}]+).*\}/i', '<?php if ($arrData[\'$1\'] $2 $3): ?>', $strTag);
-			}
-			elseif (strncmp($strTag, '{elseif', 7) === 0)
-			{
-				$strReturn .= preg_replace('/\{elseif ([A-Za-z0-9_]+)([=!<>]+)([^;$\(\)\[\]\}]+).*\}/i', '<?php elseif ($arrData[\'$1\'] $2 $3): ?>', $strTag);
+				$strReturn .= preg_replace_callback(
+					'/{(if|elseif) ([^=!<>]+)([=!<>]+)([^;$\\(\\)\\[\\]\\}]+).*\\}/i',
+					function (array $matches) use ($arrData)
+					{
+						if (!array_key_exists($matches[2], $arrData))
+						{
+							System::log(sprintf('Tried to evaluate (%s statement) unkown simple token "%s".', $matches[1], $matches[2]), __METHOD__, TL_ERROR);
+							return sprintf('<?php %s (false): ?>', $matches[1]);
+						}
+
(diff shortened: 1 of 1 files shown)


system/modules/core/library/Contao/StringUtil.php
```diff
@@ -491,13 +491,22 @@ public static function parseSimpleTokens($strString, $arrData)
...
-			elseif (strncmp($strTag, '{elseif', 7) === 0)
-			{
-				$strReturn .= preg_replace('/\{elseif ([A-Za-z0-9_]+)([=!<>]+)([^;$\(\)\[\]\}]+).*\}/i', '<?php elseif ($arrData[\'$1\'] $2 $3): ?>', $strTag);
+				$strReturn .= preg_replace_callback(
+					'/{(if|elseif) ([^=!<>]+)([=!<>]+)([^;$\\(\\)\\[\\]\\}]+).*\\}/i',
+					function (array $matches) use ($arrData)
+					{
+						if (!array_key_exists($matches[2], $arrData))
+						{
+							System::log(sprintf('Tried to evaluate (%s statement) unkown simple token "%s".', $matches[1], $matches[2]), __METHOD__, TL_ERROR);
```
xchs: Typo: "unkown" → "unk**n**own"

system/modules/core/library/Contao/StringUtil.php
```diff
@@ -491,13 +491,22 @@ public static function parseSimpleTokens($strString, $arrData)
...
-			if (strncmp($strTag, '{if', 3) === 0)
+			if (strncmp($strTag, '{if', 3) === 0 || strncmp($strTag, '{elseif', 7) === 0)
 			{
-				$strReturn .= preg_replace('/\{if ([A-Za-z0-9_]+)([=!<>]+)([^;$\(\)\[\]\}]+).*\}/i', '<?php if ($arrData[\'$1\'] $2 $3): ?>', $strTag);
-			}
-			elseif (strncmp($strTag, '{elseif', 7) === 0)
-			{
-				$strReturn .= preg_replace('/\{elseif ([A-Za-z0-9_]+)([=!<>]+)([^;$\(\)\[\]\}]+).*\}/i', '<?php elseif ($arrData[\'$1\'] $2 $3): ?>', $strTag);
+				$strReturn .= preg_replace_callback(
+					'/{(if|elseif) ([^=!<>]+)([=!<>]+)([^;$\\(\\)\\[\\]\\}]+).*\\}/i',
```
leofeyer: It seems you have added more slashes than required in `\\(\\)\\[\\]\\}]+).*\\}`, haven't you?

leofeyer:
> Obviously you cannot use any of these characters because they are used for comparison: `=!<>`.

What exactly does this mean?

aschempp:
`{if field_name="something"}`

You can't use `=` in the field name or we don't know where it ends…

leofeyer:
What happens if someone does anyway?

Toflar:
Exactly but that hasn't changed. It has always been like that :)

Toflar:
No useful output, just like now.

aschempp:
Maybe we could output something in dev mode? The console/contao dev page already shows unused insert tags, but we could add some sort of output? I can't imagine a reason where anyone would want a wrong insert tag.

aschempp:
Simple token I mean 😂

leofeyer:
That's why I asked.

Toflar:
Obsolete. See https://github.com/contao/core-bundle/pull/570.

PR #8450, opened 2016-08-26
"The registry already contains an instance" error in Backend::addFileMetaInformationToRequest

### LINKED ISSUE #8443: "The registry already contains an instance" error in Backend::addFileMetaInformationToRequest
If we save a content element with the type "text" and there's a function call, for example in a save_callback function which loads the parent page with 

```php
$objPage = \PageModel::findById($page_id);
```

the tl_content::storeFileMetaInformation function (defined in the tl_content dca as save_callback) fails, because it calls the Backend::addFileMetaInformationToRequest function and in this function the following code is executed:

```php
if ($objPage === null || $objPage->numRows < 1)
{
	return;
}

$objModel = new \PageModel();
$objModel->setRow($objPage->row());
$objModel->loadDetails();
```

If I use these calls instead, the problem is solved:

```php
if ($objPage === null || $objPage->numRows < 1 || !$objPage->id)
{
	return;
}

$objModel = \PageModel::findById($objPage->id);
$objModel->loadDetails();
```

aschempp:
Fixes #8443

### DIFF
--- a/system/modules/core/classes/Backend.php
+++ b/system/modules/core/classes/Backend.php
@@ -757,33 +757,27 @@ abstract class Backend extends \Controller
 		switch ($strPtable)
 		{
 			case 'tl_article':
-				$objPage = $db->prepare("SELECT * FROM tl_page WHERE id=(SELECT pid FROM tl_article WHERE id=?)")
-							  ->execute($intPid);
+				$objPage = \PageModel::findOneBy(array('tl_page.id=(SELECT pid FROM tl_article WHERE id=?)'), array($intPid));
 				break;
 
 			case 'tl_news':
-				$objPage = $db->prepare("SELECT * FROM tl_page WHERE id=(SELECT jumpTo FROM tl_news_archive WHERE id=(SELECT pid FROM tl_news WHERE id=?))")
-							  ->execute($intPid);
+				$objPage = \PageModel::findOneBy(array('tl_page.id=(SELECT jumpTo FROM tl_news_archive WHERE id=(SELECT pid FROM tl_news WHERE id=?))'), array($intPid));
 				break;
 
 			case 'tl_news_archive':
-				$objPage = $db->prepare("SELECT * FROM tl_page WHERE id=(SELECT jumpTo FROM tl_news_archive WHERE id=?)")
-							  ->execute($intPid);
+				$objPage = \PageModel::findOneBy(array('tl_page.id=(SELECT jumpTo FROM tl_news_archive WHERE id=?)'), array($intPid));
 				break;
 
 			case 'tl_calendar_events':
-				$objPage = $db->prepare("SELECT * FROM tl_page WHERE id=(SELECT jumpTo FROM tl_calendar WHERE id=(SELECT pid FROM tl_calendar_events WHERE id=?))")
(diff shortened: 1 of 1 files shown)


system/modules/core/classes/Backend.php
```diff
@@ -757,33 +757,27 @@ public static function addFileMetaInformationToRequest($strUuid, $strPtable, $in
 		switch ($strPtable)
 		{
 			case 'tl_article':
-				$objPage = $db->prepare("SELECT * FROM tl_page WHERE id=(SELECT pid FROM tl_article WHERE id=?)")
-							  ->execute($intPid);
+				$objPage = \PageModel::findOneBy(array('tl_page.id=(SELECT pid FROM tl_article WHERE id=?)'), array($intPid));
```
leofeyer: Why did you write `tl_page.id` instead of just `id`?

aschempp: https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Model.php#L821

The table name must always be specific because of potential relation joins.

leofeyer: Yes, but then what about the field names in the sub-queries?

aschempp: Sub-queries do not affect the names in main query.

system/modules/core/classes/Backend.php
```diff
@@ -801,17 +795,20 @@ public static function addFileMetaInformationToRequest($strUuid, $strPtable, $in
...
 		{
 			return;
 		}
 
-		$objModel = new \PageModel();
-		$objModel->setRow($objPage->row());
-		$objModel->loadDetails();
+		if (!$objPage instanceof \PageModel)
+		{
+			$objPage = \PageModel::findByPk($objPage->id);
```
leofeyer: I have already merged the PR, but now I noticed that we are not checking `!== null` here anymore. Can I change it to the following?

```php
	default:
		// HOOK: support custom modules
		if (isset($GLOBALS['TL_HOOKS']['addFileMetaInformationToRequest']) && is_array($GLOBALS['TL_HOOKS']['addFileMetaInformationToRequest']))
		{
			foreach ($GLOBALS['TL_HOOKS']['addFileMetaInformationToRequest'] as $callback)
			{
				if (($val = \System::importStatic($callback[0])->{$callback[1]}($strPtable, $intPid)) !== false)
				{
					$objPage = $val;
				}
			}
					
			if ($objPage instanceof \Database\Result && $objPage->numRows < 1)
			{
				return;
			}

			if (!$objPage instanceof \PageModel)
			{
				$objPage = \PageModel::findByPk($objPage->id);
			}
		}
		break;
}

if ($objPage === null)
{
	return;
}

$objPage->loadDetails();
```

aschempp: uhm, we're checking it on line 789?

aschempp: theoretically, but very unlikely, the ID could not exist. Your changes are fine to me 👍

leofeyer:
Merged in b16c5d657ff3a1d16eb5c3848eed38eda22855b1 then.

PR #8459, opened 2016-09-06
Please load langauge after getPageLayout hook

tsarma:
I am using hook getPageLayout to load language dynamically. Since language is loaded before the hook it did not work. As mention in hooks documentation "It can be used to modify the page or layout object", it should be able to change $objPage->language to new value.


leofeyer:
Yes, just execute the following code in your hook:

```php
$GLOBALS['TL_LANGUAGE'] = 'new language';
\System::loadLanguageFile('default');
```

tsarma:
Thank you, it works too.

PR #8474, opened 2016-09-11
Security hardening: Set secure cookie flag when using SSL

Whissi:
Even if a website is already accessible via HTTPS most setups provide a
fallback for HTTP users to redirect to HTTPS.

In this scenario Contao can leak sensible *_USER_AUTH cookies: Imagine a
user is in the same network like an attacker. If the attacker manages to
trick the user to access the Contao website via HTTP the attacker can
capture the user's session.

This commit mitigates the problem by activating the secure flag for cookies
when using SSL.

Tested against v3.5.16.

### DIFF
--- a/system/modules/core/classes/Frontend.php
+++ b/system/modules/core/classes/Frontend.php
@@ -549,7 +549,7 @@ abstract class Frontend extends \Controller
 		$_SESSION['TL_USER_LOGGED_IN'] = false; // backwards compatibility
 
 		// Remove the cookie if it is invalid to enable loading cached pages
-		$this->setCookie($strCookie, $hash, (time() - 86400), null, null, false, true);
+		$this->setCookie($strCookie, $hash, (time() - 86400), null, null, (\Environment::get('ssl')) ? true : false, true);
 
 		return false;
 	}
--- a/system/modules/core/classes/FrontendUser.php
+++ b/system/modules/core/classes/FrontendUser.php
@@ -190,7 +190,7 @@ class FrontendUser extends \User
 			}
 
 			// Remove the cookie if it is invalid to enable loading cached pages
-			$this->setCookie('FE_AUTO_LOGIN', $strCookie, (time() - 86400), null, null, false, true);
+			$this->setCookie('FE_AUTO_LOGIN', $strCookie, (time() - 86400), null, null, (\Environment::get('ssl')) ? true : false, true);
 		}
 
 		return false;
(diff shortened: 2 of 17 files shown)


leofeyer:
@contao/developers Is this a bugfix or a new feature?

ausi:
I think this would not be fully backwards compatible. If someone actually wants to share the cookie between HTTP and HTTPS it would not work anymore.

In favor of security it may be OK to break BC here.

But wasn’t there an issue which prevented the use of cookie secure? @Toflar has it on his list in contao/core-bundle#574.

xchs:
Related issue: https://github.com/contao/core-bundle/issues/443

Whissi:
Yes, it will break sites which serves HTTP and HTTPS at the same time.
Well, it doesn't really break something, it just doesn't allow you to downgrade.

I.e. when coming from HTTPS with HTTPS-only Cookie and downgrade to HTTP then the HTTPS-only Cookie is missing and the "session" looks like a fresh session (i.e. you would have to re-login and stuff like that).
If you are coming from HTTP and upgrade to HTTPS this won't be a problem.

I don't consider that as an issue. For me a site which allows HTTP-downgrade has an improper configuration (this site will for example also have a duplicated content problem because search engines like Google treats HTTP and HTTPS as different sites).

leofeyer:
@contao/developers Bug or feature?

ausi:
IMO Feature

aschempp:
Feature

discordier:
Feature but I assume an obsolete one with the auth PR by @bytehead.

Toflar:
It is a feature but a very important one. That's why I already started working on it and unfortunately there are still show stoppers. See task (and prerequisite) number 5 in https://github.com/contao/core-bundle/issues/574. If this blocker is out of the way we can even support both, page tree's using https and page tree's using http in the same installation.

leofeyer:
While porting the changes to Contao 4.3, I noticed another scenario that would break.

1. Think of a multi-domain installation with one website being available via HTTPS and the others only via HTTP.
2. The option "search protected pages" is enabled.
3. The back end is available via HTTPS, too.
4. In the maintenance module, the search index is rebuilt with a front end user logged in.

Since the back end is called via HTTPS, the cookies will be set with the secure flag. But since only one website is available via HTTPS, the front end user would not be logged in on the other website and the protected pages would not be indexed.

aschempp:
The cookie cannot be set for a foreign domain anyway...

leofeyer:
That is true.

leofeyer:
Implemented in contao/core-bundle@05e725a632aea8662896da18ad18b76a0f079bf8.

PR #8527, opened 2016-10-14
[RTM] Fixed Controller::getForm()

Toflar:
which was relying on the Form class and always being rendered as a content element.

see https://github.com/contao/core-bundle/pull/582

### DIFF
--- a/system/modules/core/library/Contao/Controller.php
+++ b/system/modules/core/library/Contao/Controller.php
@@ -507,10 +507,11 @@ abstract class Controller extends \System
 	 *
 	 * @param mixed  $varId     A form ID or a Model object
 	 * @param string $strColumn The column the form is in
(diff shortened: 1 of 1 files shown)


aschempp:
Why not just pass in the FQCN?

Toflar:
Because it does not make sense? You can only render it as CE or MOD. There is no other possibility.

aschempp:
somewhat true, yes. But it would be some sort of *constant* instead of a strange string.

Toflar:
¯\_(ツ)_/¯

leofeyer:
As discussed in Mumble on December 14th, @Toflar will change the argument to `$blnModule` and then the PR can be merged.

Toflar:
Updated.

aschempp:
Personally I prefer true flags to enable a new feature instead of this, but obviously both works.

leofeyer:
Merged in contao/core-bundle@078cb3438f212b6f5a5d722ea7707851e2110bc7.

Issue #7660, opened 2015-02-26
The getForm hook


dmolineus:
The getForm hook is not triggered when including the form as content element or module. It's because the form is rendered as Hybrid. The `Controller::getForm` is never called.

I propose to move the hook into `Form::generate` so it would be possible to use the hook for the common way how forms are included into Contao.

leofeyer:
What exactly do you want to accomplish? You might as well use the "getContentElement" or "getFrontendModule" hook to modify the form object.

dmolineus:
Depending on the form settings I have to load some frontend validations. Doing a workaround with the proposed `getContentElement` or `getFrontendModule` hooks is definitly possible.

The question for me is if this hook is intended to be called for every form which is loaded or just for the forms which are loaded by the controller.

dmolineus:
Checked. Still relevant.

aschempp:
Where would `Controller::getForm` be used if not for modules and content elements?

dmolineus:
> Where would Controller::getForm be used if not for modules and content elements?

I don't understand? The point is that the getForm hook is not triggered when `Controller::getForm` is called.

PR #8529, opened 2016-10-17
Move getForm hook into the form hybrid.

dmolineus:
As stated here #7660 the `getForm` hook is only triggered then using `Controller::getForm` and not when a form content element is rendered. I suggest to move the hook into the `Form` class to fix this issue.


leofeyer:
There are also hooks in the `Controller::getFrontendModule()` and `Controller::getContentElement()` methods, which are not triggered in the `generate()` methods of the corresponding classes. Therefore I am against merging this.

dmolineus:
The difference is that for the regular use case (using content element in an article) the `getContentElement` method is triggered. Same for frontend modules in the regular page.

https://github.com/contao/core/blob/9ca7b0d03b522622ce4bd2b00ad4fc34efbf28f6/system/modules/core/modules/ModuleArticle.php#L213
https://github.com/contao/core/blob/9ca7b0d03b522622ce4bd2b00ad4fc34efbf28f6/system/modules/core/pages/PageRegular.php#L133-L137

So the hooks are called in the default way Contao works. Only the getForm hooks is only triggered when a form is used as insert tag or a third party extension calls `Controller::getForm`.

leofeyer:
But in these cases, the `getContentElement` or `getFrontendModule` hooks are triggered. Then the `getForm` hook would be triggered additionally, wouldn't it?

dmolineus:
> But in these cases, the getContentElement or getFrontendModule hooks are triggered. Then the getForm hook would be triggered additionally, wouldn't it?

Sorry, missed that question up to now.

No, that's not the case. Because than the hybrid class is instantiated and not the `Controller::getForm` is called.

ausi:
Moving the hook would be a BC break because it might not be expected that the hook is called for every form.

We would have to add an additional hook and I don’t know if we want that either.

Toflar:
`Controller::getForm()` works as a factory here so the issue is rather that the content element is not using `getForm()`. This should be the fix, imho.

ausi:
You should probably use the existing hooks for getContentElement and getFrontendModule instead and check there if the element is a form.

dmolineus:
> Moving the hook would be a BC break because it might not be expected that the hook is called for every form.

Also it might not be expected that the hook is **not** called. The docs claims it allows to manipulate the generated form: https://docs.contao.org/books/api/extensions/hooks/getForm.html#getform

> Controller::getForm() works as a factory here so the issue is rather that the content element is not using getForm(). This should be the fix, imho.

Would be fine with me.

> You should probably use the existing hooks for getContentElement and getFrontendModule instead and check there if the element is a form.

That's what I ended up doing here. But still think, it's wrong. :D

Toflar:
> Would be fine with me.

Could you update the PR accordingly then? That should fix the issue, right?

dmolineus:
>> Would be fine with me.
>
> Could you update the PR accordingly then? That should fix the issue, right?

Hm, had a look in the code again. My answer was false. It would end in an recursive loop as `Controller::getForm` instantiates the `Form` hybrid. If `Form::generate` calls `Controller::getForm` your webserver will be happy. :)

leofeyer:
@dmolineus If you still need this, please create the PR again at https://github.com/contao/contao.

PR #8530, opened 2016-10-17
Wrap combined file content with global namespace

dmolineus:
Since Contao supports callables for dca callbacks and PHP 5.5 introduced the `Class:class` constant there are valid cases using the `use statement` for namespaced classes in the dca and config files.

At the moment you cannot use the `use statement` because of the danger that the combined files could have duplicate use statements. This pull request fixes it by wrapping each file content in a namespace notation.

So it would get possible to use following code safely:

**DCA file**

``` php
<?php 
// dca/tl_content.php
use Vendor\Namespace\GetTemplatesCallback;

$GLOBALS['TL_DCA']['tl_content']['fields']['customTemplate']['options_callback'] = new GetTemplatesCallback('ce_custom');
```

**Config file**

``` php
<?php 
// config/config.php
use Vendor\Namespace\CustomElement;
$GLOBALS['TL_CTE']['text']['custom'] = CustomElement::class;
```

### DIFF
--- a/system/modules/core/library/Contao/Automator.php
+++ b/system/modules/core/library/Contao/Automator.php
@@ -494,7 +494,9 @@ class Automator extends \System
 
 			if (file_exists(TL_ROOT . '/' . $strFile))
 			{
-				$objCacheFile->append(static::readPhpFileWithoutTags($strFile));
+				$strBuffer = static::readPhpFileWithoutTags($strFile);
+				$strBuffer = $this->wrapWithGlobalNamespace($strBuffer);
+				$objCacheFile->append($strBuffer);
 			}
 		}
 
@@ -599,7 +601,9 @@ class Automator extends \System
 
 				if (file_exists(TL_ROOT . '/' . $strFile))
 				{
-					$objCacheFile->append(static::readPhpFileWithoutTags($strFile));
+					$strBuffer = static::readPhpFileWithoutTags($strFile);
+					$strBuffer = $this->wrapWithGlobalNamespace($strBuffer);
+					$objCacheFile->append($strBuffer);
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


leofeyer:
What is wrong with `$GLOBALS['TL_CTE']['text']['custom'] = 'Vendor\Namespace\CustomElement';`?

aschempp:
@leofeyer that's a different case, you're referring to a string which always must be used with FQCN, but @dmolineus is referring to calling code in the respective files.

leofeyer:
I wanted to point out that it is not necessary to use `use` statements in DCA files. As we all agree, DCA files should not have contained classes and functions in the first place, therefore it seems wrong to me to implement stuff that further encourages developers to do so.

Our goal should be to reduce PHP usage in DCA files, so we can e.g. use YAML files in the future.

dmolineus:
> What is wrong with $GLOBALS['TL_CTE']['text']['custom'] = 'Vendor\Namespace\CustomElement';?

Nothing wrong with that. The benefit of `CustomElement::class` would be that renaming or moving the class the reference would automatically updated. 

And yes, the class does not have to be loaded to use the `::class`. In PHP `::class` also works for non existing classes.

dmolineus:
> I wanted to point out that it is not necessary to use use statements in DCA files. As we all agree, DCA files should not have contained classes and functions in the first place, therefore it seems wrong to me to implement stuff that further encourages developers to do so.

Using the `::class` notation is a very useful feature. Allowing use statements would increase the readability of the dca files when using multiple callback definitions. 

Furthermore Contao supports callable here since 3.2(?). For using class names inside anonymously functions using the use statement would be helpful.

> Our goal should be to reduce PHP usage in DCA files, so we can e.g. use YAML files in the future.

As long as the PHP are used, standard PHP features should be supported as well. Detecting name refences by IDEs when renaming or moving classes is one benefit from it (and a drawback of YAML files).

Toflar:
I'm +1 on this PR.

discordier:
I am also :+1:

I have experienced the described issue in the past and had to remove the imports from `config.php` files.

PR #8543, opened 2016-10-26
Also check for file existence in example for URLs without any suffix

Toflar:


.htaccess.default
```diff
@@ -258,6 +258,7 @@ FileETag None
   # If you do not want to use an URL suffix at all, you have to add a third
   # line to prevent URLs that point to folders from being rewritten (see #4031).
   #
+  #   RewriteCond %{REQUEST_FILENAME} !-f
```
leofeyer: This code is already present in line 273, so no need to add it here.

Toflar: Ah, I didn't read properly. It says "add a third line". I replaced it. Never mind.

PR #8544, opened 2016-10-26
Cache fallback pages

Toflar:
I noticed that `PageModel::findWithDetails()` executes a lot of very identical queries. I figured it was the query to fetch the fallback page. As you can see, whenever you call `findWithDetails()`, Contao works up the tree and calls the fallback page for every page again, even though the information is already there. I inspected an Isotope installation where there was some heavy querying going on because of that. To give you some numbers: With this fix, the number of queries dropped from 1121 for a page to 394 ;-) So I consider this a quite important bugfix that should go into 3.5.

### DIFF
--- a/system/modules/core/models/PageModel.php
+++ b/system/modules/core/models/PageModel.php
@@ -723,7 +723,7 @@ class PageModel extends \Model
 	/**
 	 * Register the contao.dns-fallback alias when the model is attached to the registry
 	 *
-	 * @param \Model\Registry|\Contao\Model\Registry $registry The model registry
+	 * @param \Model\Registry $registry The model registry
 	 */
 	public function onRegister(\Model\Registry $registry)
 	{
@@ -740,7 +740,7 @@ class PageModel extends \Model
 	/**
 	 * Unregister the contao.dns-fallback alias when the model is detached from the registry
 	 *
-	 * @param \Model\Registry|\Contao\Model\Registry $registry The model registry
+	 * @param \Model\Registry $registry The model registry
 	 */
 	public function onUnregister(\Model\Registry $registry)
 	{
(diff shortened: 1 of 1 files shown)


leofeyer:
Agreed. Nevertheless, the query should already be cached by MySQL, so I don't think this will have that big an impact.

Toflar:
727 Queries times 0.3 ms. You do the math yourself. :)

leofeyer:
You are right, the impact is huge! Very good find! I have adopted the fix in contao/core-bundle@f9d926df8cc525a4b7aa8969ed144c38b010439f for version 4.2.5. Will also merge the PR for Contao 3.

aschempp:
Totally agree, but one thing is flawed: The array must only contain the ID of the model, not the actual object. And retrieve it from the registry using a `findByPk` call.

Toflar:
Updated to use the registry for that stuff (which I was actually building the API for when improving fetching from unique fields, dunno how I could forget about that :D)

system/modules/core/models/PageModel.php
```diff
@@ -710,6 +721,40 @@ public static function findWithDetails($intId)
...
+	 * Called when the model is attached to the model registry
+	 *
+	 * @param \Model\Registry|\Contao\Model\Registry $registry The model registry
+	 */
+	public function onRegister(\Model\Registry $registry)
+	{
+		parent::onRegister($registry);
+
+		// Register this model as being the fallback page for a given dns
+		if ($this->fallback && !$registry->isRegisteredAlias($this, 'dns-fallback', $this->dns))
```
aschempp: it should never be possible that there is already an alias for that value?

aschempp: Additionally, you're not checking if this is the root page.

leofeyer:
Merged in 8e9bdd77fa4160026fe5e225c5f5af3b1973c3b6.

PR #8545, opened 2016-10-26
Unlock members after password change (fixes #8542)

### LINKED ISSUE #8542: Login right after password change may fail
Today we launched a site which has thousands of users imported from the old page. The users has been informed about the relaunch via newsletter so we currently got plenty of traffic on it. Because we could not decrypt the old passwords all users now have to request a new one - nothing unusual so far. But many users are now complaining that their new passwords don't work. We can confirm this. Right after setting a new password the login might fail from time to time.

I dug a bit into the core and noticed that the User class does a simple `execute` for it`s SELECT query which might lead into cached results.

My idea: Users try to login using their old password (which of course fails), request a new one and try to login again. But because the old SELECT statement is still cached we get the old password and the login fails.

So I think the User class should use `executeUncached` instead of `execute`. What are your thoughts on this? The line in question is this one https://github.com/contao/core/blob/master/system/modules/core/library/Contao/User.php#L527

discordier:


leofeyer:
Ported the changes for the release of version 4.2.5 in contao/core-bundle@11e79972335104f1a054b25e749054dcbbd5601d.

leofeyer:
Merged in 05e96c98d0c6d3cfd4f5809f5af05bf9a2ac9704.

PR #8557, opened 2016-11-03
fix typo in Widget comment

fritzmg:
1. there was a typo in the comment saying that the `extnd` regex disallows `/`, but it actually disallows `\`
2. the order of the symbols is now equal to [Validator::isExtendedAlphanumeric](https://github.com/contao/core/blob/3.5.18/system/modules/core/library/Contao/Validator.php#L95-L105)

Noticed via #8556

### DIFF
--- a/system/modules/core/library/Contao/Widget.php
+++ b/system/modules/core/library/Contao/Widget.php
@@ -967,7 +967,7 @@ abstract class Widget extends \BaseTemplate
 					}
 					break;
 
-				// Do not allow any characters that are usually encoded by class Input [=<>()#/])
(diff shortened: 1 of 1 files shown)


leofeyer:
Merged in 4d75de5f752dd7f683f6bb24fe7ccb1594e207e9.

leofeyer:
Just for the record: The comment in the `Validator` class is much clearer than the one in the `Widget` class.

```
Characters that are usually encoded by class Input: #<>()\=
```

```
Do not allow any characters that are usually encoded by class Input ([#<>()\=])
```

Are the square brackets in the second comment part of the disallowed characters or is it a regex character class?

PR #8561, opened 2016-11-07
Adding method findOne to Model.php

SyZn:
The `function find` is **protected** and it cannot be used from outside the Model context. The function `findAll` on the other hand is declared **public** and can be used to search for all occurrences but there is no possiblity (or maybe I overlooked it?) to search for 2 or more fields (e.g. pid and module) in a Standard model like ContentModel and get just **one result**. (I would have to override the Model class just to do some migration steps) So I propose this additional function in the Model.php to enhance the possibilities to query the Database using Models.

### DIFF
--- a/system/modules/core/library/Contao/Model.php
+++ b/system/modules/core/library/Contao/Model.php
@@ -980,6 +980,29 @@ abstract class Model
 
 		return static::find($arrOptions);
 	}
+	
+	/**
+	 * Find the first record
+	 * 
+	 * @param array $arrOptions An optional options array
+	 * 
+	 * @return static|null The Model or null if the result is empty
+	 */
+	public static function findOne(array $arrOptions=array())
+	{
+		$arrOptions = array_merge
+		(
+			array
+			(
+				'limit'	 => 1,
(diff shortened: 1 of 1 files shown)


bezin:
This should be achieved by using the findOneBy method in the model class, shouldn't it?

fritzmg:
Yes, just use `findOneBy`. `$strColumn` and `$varValue` can be arrays as well.

SyZn:
I've tried that one but it did not work out for me, but maybe it was just a bug somewhere else. I will try again and post the results here.
What should be in the `findOneBy` for the `$strColumn`: `array('pid', 'name')` or `array('pid=?', 'name=?')`. Btw. it would be great if in that case `strColumn`would be called `varColumn`.

fritzmg:
It should be
```php
$t = \MyModel::getTable();
$record = \MyModel::findOneBy(array("$t.pid = ?", "$t.name = ?"), array(…, …));
```

aschempp:
Both `$strColumn` and `$strValue` accept array on all the model methods.

PR #8562, opened 2016-11-07
Do not convert paths to uuid for paths outside of the upload path.

dmolineus:
This PR fixes the issue that values get lost when using the file selector for paths outside of the upload path (dbafs).

### DIFF
--- a/system/modules/core/widgets/FileSelector.php
+++ b/system/modules/core/widgets/FileSelector.php
@@ -456,6 +456,12 @@ class FileSelector extends \Widget
 			return;
 		}
 
+		// Do not convert values if path is not in the upload path
+		if ($this->path !== \Config::get('uploadPath'))
+		{
+			return;
+		}
+
 		// Ignore the numeric IDs when in switch mode (TinyMCE)
 		if (\Input::get('switch'))
 		{
--- a/system/modules/core/widgets/FileSelector.php
+++ b/system/modules/core/widgets/FileSelector.php
@@ -457,7 +457,7 @@ class FileSelector extends \Widget
 		}
 
 		// Do not convert values if path is not in the upload path
-		if ($this->path !== \Config::get('uploadPath'))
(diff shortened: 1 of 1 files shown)


system/modules/core/widgets/FileSelector.php
```diff
@@ -456,6 +456,12 @@ protected function convertValuesToPaths()
 			return;
 		}
 
+		// Do not convert values if path is not in the upload path
+		if ($this->path !== \Config::get('uploadPath'))
```
dmolineus: Afaik the path does not store the value, but the eval option path. This is actually how it works in my environment as well.

leofeyer: Ah, I see. Still, what if `$this->path` is a subfolder of `Config::get('uploadPath')`? We must only return if the path is outside the upload directory IMHO.

dmolineus: Valid point. Improved the condition also handling sub path of the upload path.

system/modules/core/widgets/FileSelector.php
```diff
@@ -457,7 +457,7 @@ protected function convertValuesToPaths()
 		}
 
 		// Do not convert values if path is not in the upload path
-		if ($this->path !== \Config::get('uploadPath'))
+		if (strncmp($this->path, \Config::get('uploadPath'), strlen(\Config::get('uploadPath'))) !== 0)
```
fritzmg: Why not
```php
if (strpos($this->path, \Config::get('uploadPath')) === false)
```
?

leofeyer: To make it 100% accurate, it should be:

```php
if ($this->path != \Config::get('uploadPath') && strpos($this->path, \Config::get('uploadPath').'/') !== 0)
```

Otherwise the condition will evaluate to `true` if `$this->path` is e.g. `files-and-docs`.

leofeyer:
Fixed in 7a2edc6f929af901dac6ee6abb05cd17a8f347c0.

Issue #8558, opened 2016-11-03
mediaelement.js and picture element problem on IE9


may17:
The picture default template uses an conditional comment for ie9 wich inserts an video tag. This video tag is also interpreted by mediaelement.js on ie9 and an error occurs because there is no valid sources. This issue brings up horrible results on pages with images ;)

To fix this issue it would be better to set the selector in the j_mediaelement file to .ce_player video and .ce_youtube video.

Even if IE9 support has no high priority it would be generally better to concrete the selector to avoid further problems.

leofeyer:
You mean

```js
$('.ce_player video,.ce_youtube video,audio').mediaelementplayer({
```

instead of

```js
$('video,audio').mediaelementplayer({
```

?

may17:
Yes looks great. Thank You.

may17:
I again slept on it and perhaps the solution:

`$('.ce_player video,.ce_youtube video,audio').mediaelementplayer({`

brings up problems in some systems because we only match for ce_player and ce_youtube but there are a lot of third party extensions and customizations out there that might be broken with this fix.

We need a selector like "invoke mediaelement on every video without video tags in picture tags".

Something like this:

```javascript
var media_items = $('video,audio').filter(function() {
  if($(this).parent().prop("tagName") !== "PICTURE") {
     return this;
  }
});

media_items.mediaelementplayer({
```

leofeyer:
That is even better IMHO. Can we also apply it to `'video,audio'`?

may17:
Yes we can. I will create an pull request for this issue.

leofeyer:
I'm closing this in favor of #8568.

PR #8568, opened 2016-11-10
Fix for ticket #8558

may17:


system/modules/core/templates/jquery/j_mediaelement.html5
```diff
@@ -9,7 +9,12 @@ $GLOBALS['TL_CSS'][] = 'assets/jquery/mediaelement/'. $GLOBALS['TL_ASSETS']['MED
 <script>
   (function($) {
     $(document).ready(function() {
-      $('video,audio').mediaelementplayer({
+
+      var media_items = $('video').filter(function() {
+        return $(this).parent().prop("tagName") !== "PICTURE";
```
leofeyer: Shouldn't this be `.tagName.toLowerCase() !== 'picture'`? I don't think we need to use `prop()` here.

may17: tagName is no method of jquery and parent returns an jQuery element. So this only works when we use plain javascript but jquery is required so why we shouldn`t use it? Even the filter method is jQuery too :)

See this jsbin:
https://jsbin.com/toxicemoqi/edit?html,js,console,output

Prop returns a string and we could lowercase the value if you like.

may17: And yes we could return the plain javascript element with .parent()[0] or .parent().get(0).

Looks like this: $(this).parent().get(0).tagName.toLowerCase() !== 'picture'

Or in vanilla js:

this.parentElement.tagName.toLowerCase();

leofeyer: Let's just use `prop()`. 😄

aschempp (approved):
Looks good, though I would keep `$('video,audio')` to not create a new variable. Won't harm to filter the audio elements as well.

system/modules/core/templates/jquery/j_mediaelement.html5
may17: +1 I think thats the cleanest way @aschempp

may17:
@aschempp it looks cleaner for me only to filter first the video items cause thats our goal.

I think you mean something like this:
```javascript
$('video,audio').filter(function() {
    return $(this).parent('picture').length > 0;
}).mediaelementplayer({
```

Both is possible it`s just about personal taste :)

aschempp:
Yeah, though it probably must be reversed, right? The parent *must not* be a picture element 😂

aschempp:
So best probably would be this?

```js
$('video,audio').filter(function() {
    return $(this).parents('picture').length === 0;
}).mediaelementplayer({
```

This will make sure it also works if there is another element around the video, for whatever reason.

may17:
I`ve updated my pull request. I think we`ve found a good solution :)

aschempp (approved):

leofeyer:
Merged in b6c25522eedc992e30326717d8131aef449ba3ff and updated the XHTML template in 95340c455a010b2f8fbe584e00836e1a7a9b73c8.

PR #8570, opened 2016-11-15
Added a simple event details query cache

Toflar:
The event class checks if an event has details for every single event. Especially when using repeating events, this causes a lot of unnecessary database queries.
In the Contao demo we have 3 events that are repeated. This simple PR reduced my DB queries from 130 down to 46.

### DIFF
--- a/system/modules/calendar/classes/Events.php
+++ b/system/modules/calendar/classes/Events.php
@@ -43,6 +43,12 @@ abstract class Events extends \Module
 	 */
 	protected $arrEvents = array();
 
+	/**
+	 * Has details cache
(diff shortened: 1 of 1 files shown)


aschempp:
Why's the cache not in the model?

Toflar:
Because it is wrong. We could add caches to every count method of every model then. We removed the db query cache because it caused troubles. The developer has to know whether the result can be cached, not the model.

leofeyer:
As briefly discussed, I have implemented lazy loading in 8cc5aa8689475e17850a96c739fd3c24f85363a4.

PR #8574, opened 2016-11-18
[RTM] Fix the mandatoriness validation of ImageSize widget

qzminski:
Without this fix if you set the field to be mandatory and choose the predefined image size value, you will always get an error as ```$varInput[1]``` will be empty.

### DIFF
--- a/system/modules/core/widgets/ImageSize.php
+++ b/system/modules/core/widgets/ImageSize.php
@@ -76,7 +76,7 @@ class ImageSize extends \Widget
 
 		if (!is_numeric($varInput[2]))
 		{
-		    $varInput[0] = parent::validator($varInput[0]);
+			$varInput[0] = parent::validator($varInput[0]);
 			$varInput[1] = parent::validator($varInput[1]);
 		}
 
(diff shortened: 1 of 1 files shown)


leofeyer:
Your fix does not cover all use cases:





Generally, there are options which do not require width and height, there are options wich require either width or height and there are options which require both width and height.

qzminski:
What do you propose then? Have an individual check for every non-predefined crop mode?

leofeyer:
Exactly.

qzminski:
I'll update the code when I find some spare minutes then.

qzminski:
Hmm, so we have two different validation steps here:

1. If field is mandatory the user must choose the image size value from the select dropdown.
2. Despite the field mandatoriness (!) if user chose the image size, the width/height inputs should be validated. It's wrong if one selects ```center_center``` and doesn't enter any dimensions.

And to gather all modes and their validation:

1. Predefined size - no width/height validation, irrelevant
2. Relative dimensions group - must have at least width/height
3. Exact dimensions group - must have both width/height

Am I right? Can you confirm before I proceed with the code?

leofeyer:
You are absolutely right.

Toflar:
Closed due to 3.5.20 being merged. Please reopen and switch target branch to `master`.

qzminski:
I have added the validation of the relative and exact dimensions in ce16875, please review.

leofeyer:
Looks good to me. Thank you, @qzminski.

leofeyer (approved):

PR #8575, opened 2016-11-21
Modules permission when included as Content-Element

tsarma:
Check for access permission of a module when placed as Content Element of type Module.

### DIFF
--- a/system/modules/core/elements/ContentModule.php
+++ b/system/modules/core/elements/ContentModule.php
@@ -44,6 +44,11 @@ class ContentModule extends \ContentElement
 		{
 			return '';
 		}
+		
+	        if (!\Controller::isVisibleElement($objModule))
+        	{
+	            return '';
+        	}		
 
 		$objModule->typePrefix = 'ce_';
 


aschempp:
A module does not have publishing permissions? Or do you want to check the group rights?

fritzmg:
The included module might have access rights. By adding 
```php
if (!\Controller::isVisibleElement($objModule))
{
    return '';
}
```
those access rights will be considered as well.

tsarma:
@aschempp Yes you are right I mean member group rights here, I just used the term permission loosely here.

leofeyer:
Access rights are intentionally not inherited from an alias element. You need to protect the include element instead (if supported).

tsarma:
OK, if thats intentional. But I see that as a limitation.

leofeyer:
I don't. It allows you to re-use content with different access permissions.

tsarma:
> It allows you to re-use ```content``` with different access permissions.

```content``` here being an aliased module. So different access permissions can be achieved by duplicating the module with different access permissions.

I agree, with your disagreement of my view to it as limitation. I see on top of that it is also an inconsistency. Please let me try to explain why I am thinking so.

1- Limitation: Say you have placed a module on many pages on your site using insert module content element. Later on you decided to restrict that module to certain member group. Now you have to go through all those content element and set access permissions. This, I see as limitation. Instead of being able to change permission at one place, your have to change a lot of places. Also you leave a room for error.

2- Inconsistent: If there is a setting in a module (now in this case modules Access protection). User expect to work it through out the system.

PR #8587, opened 2016-12-06
Added attribute 'selected' in module Quicknav

ghost:
I added a simple feature to the quicknav module to optimize usability:
If one of the pages listed in the quicknav module is the current page, the html attribute 'selected' will be added to the < option > tag. 
 
Example:
In this example 'Dual Top Schrauben' is the current page. 
If I reload the page, 'Dual Top Schrauben' is already preselected.
bildschirmfoto 2016-12-06 um 18 27 50

### DIFF
--- a/system/modules/core/modules/ModuleQuicknav.php
+++ b/system/modules/core/modules/ModuleQuicknav.php
@@ -154,6 +154,7 @@ class ModuleQuicknav extends \Module
 				{
 					$arrPages[] = array
 					(
+						'active' => ($objPage->id == $objSubpage->id),
 						'level' => ($level - 2),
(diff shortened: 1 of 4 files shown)


asaage:
This makes "Kategorie wählen" more or less obsolete and imo reduces the navigational aspect of this module.
```<option class="active">...``` would be nice though but lacking the ability to style that pretty much useless as well.

comolo:
In cases the current page is not listed in the pagelist of the quicknav module, there is still a default text needed.

leofeyer:
Implemented in contao/core-bundle@0f4bb715f7200a4e8d592b11dd35e803c802f88d.

PR #8589, opened 2016-12-07
[RFC] Use the OpenSSL instead of the mcrypt extension

leofeyer:
This PR replaces the mcrypt implementation with an OpenSSL implementation in the `Encryption` class as discussed in #8534.

The new implementation works, however, it does not decrypt existing data that has been encrypted by `mcrypt`. We would have to decrypt and re-encrypt the data somehow.

@contao/developers What do you think?

### DIFF
--- a/system/config/default.php
+++ b/system/config/default.php
@@ -171,16 +171,10 @@ $GLOBALS['TL_CONFIG']['ftpPort'] = 21;
  * Please provide an encryption key that is at least 8 characters long. Note
  * that encrypted data can only be decrypted with the same key! Therefore
  * note it down and do not change it if your data is encrypted already.
- *
- *   encryptionMode   = defaults to "cfb"
- *   encryptionCipher = defaults to "rijndael 256"
- *
- * See PHP extension "mcrypt" for more information.
  */
-$GLOBALS['TL_CONFIG']['encryptionKey']    = '';
-$GLOBALS['TL_CONFIG']['encryptionMode']   = 'cfb';
-$GLOBALS['TL_CONFIG']['encryptionCipher'] = 'rijndael-256';
-$GLOBALS['TL_CONFIG']['bcryptCost']       = 10;
+$GLOBALS['TL_CONFIG']['encryptionKey'] = '';
+$GLOBALS['TL_CONFIG']['cipherMethod']  = 'AES-256-CBC';
+$GLOBALS['TL_CONFIG']['bcryptCost']    = 10;
 
 
 /**
--- a/system/modules/core/library/Contao/Encryption.php
+++ b/system/modules/core/library/Contao/Encryption.php
@@ -33,12 +33,6 @@ class Encryption
 	 */
 	protected static $objInstance;
 
-	/**
-	 * Mcrypt resource
-	 * @var resource
-	 */
-	protected static $resTd;
-
 
 	/**
 	 * Encrypt a value
@@ -65,22 +59,16 @@ class Encryption
 			return '';
 		}
 
-		// Initialize the module
-		if (static::$resTd === null)
-		{
-			static::initialize();
-		}
+		static::checkRequirements();
 
 		if (!$strKey)
 		{
 			$strKey = \Config::get('encryptionKey');
 		}
 
-		$iv = mcrypt_create_iv(mcrypt_enc_get_iv_size(static::$resTd));
-		mcrypt_generic_init(static::$resTd, md5($strKey), $iv);
-		$strEncrypted = mcrypt_generic(static::$resTd, $varValue);
+		$iv = openssl_random_pseudo_bytes(openssl_cipher_iv_length(\Config::get('cipherMethod')));
+		$strEncrypted = openssl_encrypt($varValue, \Config::get('cipherMethod'), $strKey, OPENSSL_RAW_DATA, $iv);
 		$strEncrypted = base64_encode($iv.$strEncrypted);
-		mcrypt_generic_deinit(static::$resTd);
 
 		return $strEncrypted;
 	}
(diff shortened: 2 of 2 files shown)


aschempp:
Why don't we use a well-established encryption library and only upgrade Contao 4?

leofeyer:
First of all, I guess that OpenSSL is a well-established encryption library.

You are probably thinking of an abstraction layer such as phpseclib, which is built upon OpenSSL? Keep in mind that the OpenSSL extension is a requirement to run Contao, therefore we will not need the native implementations of phpseclib (unnecessary overhead).

Second of all, any other encryption library will most likely cause the same problem: not being able to decrypt the already encrypted data. And this is the bigger problem IMO.

leofeyer:
Just for reference, here is an article which explains the cipher incompatibilities:

http://thefsb.tumblr.com/post/111035508040/porting-php-code-from-mcrypt-to-openssl

Sadly, OpenSSL does not handle `RIJNDAEL-256-CFB`, which we have been using so far.

leofeyer:
Before we are talking about implementations, we should answer the following two questions:

1. Do we want to fix this in Contao 3.5 or only in Contao 4?

2. Do we want to stick with `RIJNDAEL-256-CFB` or do we want to use a different method such as `AES-256-CBC` in the future?

@contao/developers What is your answer to the two questions?

Toflar:
First of all, this PR is a major BC break. It breaks all of the existing encrypted data because it will automatically try to decrypt the data using `AES-256-CBC`. We have to migrate using some technique like when upgrading the password hashing (like prefixing the encrypted data by `<cipher>:` or so whereas no cipher at all would mean `RIJNDAEL-256-CFB` if it was not changed.

Answering your two questions:

> Do we want to fix this in Contao 3.5 or only in Contao 4?

I would like to make sure you can still run Contao 3.5 on PHP 7.2 and onwards so if we can fix it by switching, we should.

> Do we want to stick with RIJNDAEL-256-CFB or do we want to use a different method such as AES-256-CBC in the future?

Also answering the question for "only Contao 4": I would like to deprecate the whole `Encryption` class because there are libraries you can load using Composer that are way better and unit tested. Also, the question what encryption cipher to use is not a general question one can answer. It depends on the data you want to encrypt and thus it's a decision that should be left to the developer. We do provide everything we need with the load and save_callbacks to develop such a feature with ease. I have already implemented something like this using `phpseclib` because I wanted to encrypt the data using the twofish algorithm. It was a matter of registering two callbacks each consisting of like 3-4 LOC so no big deal, really. Here you go:

```php
    public function encryptCallback($value)
    {
        $twofish = new Twofish();
        $twofish->setKey(\System::getContainer()->getParameter('secret'));
        return base64_encode($twofish->encrypt($value));
    }

    public function decryptCallback($ciphertext)
    {
        if (null === $ciphertext) {
            return $ciphertext;
        }

        $twofish = new Twofish();
        $twofish->setKey(\System::getContainer()->getParameter('secret'));
        return $twofish->decrypt(base64_decode($ciphertext));
    }
```

discordier:
We MUST at least provide the tools which can decrypt the old stuff and encrypt the new stuff.
We MUST do this, because Contao 3.5 maintenance window (including security) lasts until 1 Jun 2018 and therefore beyond the maintenance window of PHP 7.1 (excluding security) which lasts until 1 Dec 2018.
Therefore the people will not be able to use Contao 3.5 using a supported PHP version during the security only phase.
IMO this is not acceptable.

Second, we must provide the tools also so existing stuff can safely be decrypted and to prevent data loss.

Third, I am against changing the encryption class to use new features (openssl) in this case without deprecating it. As @Toflar pointed out already, it is not a matter of the system to provide a general purpose encryption class. It would rather be a matter of providing an encryption layer by attaching a service to DC_Table (or its potential successors) and therefore decoupling everything so different encryption may get used for different tasks which are to be specified by the developer of the application and not dictated by "the system".

How to solve the current dilemma:

IMO we should implement a notation of the used encryption method, as already pointed out by @Toflar again, within the data like we did with the password encryption.
Checking for a prefix is easy and straightforward to implement.
However, we should only apply this to decryption and remove the old style encryption altogether.
This then allows for graceful migration to the new style and only enforces a Contao and PHP combination where both methods are available during a grace period.

We could also provide a step in the maintenance section to automatically loop over all dca fields which have encryption defined and if old style is in use re-encrypt the row and save using the new algorithm.

leofeyer:
As discussed in Mumble on December 14th, we want to deprecate the `'encrypt' => true` flag and the corresponding encryption methods of the `Encryption` class (not the hashing methods though) and suggest people to implement their own preferred encryption via load and save callbacks.

leofeyer:
Deprecations added in d1157c528c4286b4e664457f02f94b73628298cc.

aschempp:
we should also add deprecation notices to `eval => encrypted = true` shouldn't we?

leofeyer:
Then there would be two deprecation notices, wouldn't there?

aschempp:
True, but otherwise you would not notice where it came from, right?

leofeyer:
Hm … Where would you add the `trigger_error()`? There are quite a few usages.

aschempp:
In DC_Table ?

leofeyer:
Certainly not. Maybe in `Widget::getAttributesFromDca()`, but IMO it is not worth the overhead.

discordier:
Well, it is the only place where this is centrally checked.

IMO we should announce deprecations in there.
We also could provide a dca-checker class which validates DCA array keys and values against known constraints. Pretty much like `composer validate`.

aschempp:
> We also could provide a dca-checker class which validates DCA array keys and values against known constraints.

Yeah we could… I don't believe that's ever gonna happen 😂

leofeyer:
Changed in 5d59f95aa9342e9682e14937f10365352fbb93d5.

PR #8597, opened 2016-12-15
[RTM] Support cross domain requests when rebuilding the search index

Toflar:
This is a PR for https://github.com/contao/core-bundle/issues/633.

Due to the fact that we are requesting using ajax, it's never a simple request CORS case (header `X-Requested-With`) so you should read up upon the general CORS concept and preflighted requests.

This PR

* makes sure CORS headers are only set when a `GET` request is sent together with the `Contao-Rebuild-Search-Index` header.
* for maximum security, the `Access-Control-Allow-Origin` header is only set when there's also a root page responsible for that origin.
* also sets the headers on redirect responses which is currently not supported in the major browsers.

As of today it does not work for pages in the search index that send 3xx redirect headers because the initial standard forbid redirects in preflighted requests. This [has been changed](https://github.com/whatwg/fetch/commit/0d9a4db8bc02251cc9e391543bb3c1322fb882f2) only a few months ago and is not adopted in the browsers yet. Chromium seems to be [about to merge support for the adjusted spec](https://bugs.chromium.org/p/chromium/issues/detail?id=580796).

I ran into that issue because Isotope sends a redirect in the filter module (using `Controller::redirect()`) so you as of today you end up having something like this:

bildschirmfoto 2016-12-15 um 14 53 14

So the PR is already prepared and does send the correct headers on redirect responses too but until the browser vendors are adjusting to the updated spec, there might be a few pages failing. Anyway, I think we can live with that as there will be only a little number of pages that fail. :-)

I would like to thank @netzarbeiter for setting up a test environment for me right after we had the Contao call. Open source work at its best 👍

### DIFF
--- a/system/modules/core/library/Contao/Controller.php
+++ b/system/modules/core/library/Contao/Controller.php
@@ -996,26 +996,7 @@ abstract class Controller extends \System
 	 */
 	public static function reload()
 	{
-		if (headers_sent())
-		{
-			exit;
-		}
-
-		$strLocation = \Environment::get('uri');
-
-		// Ajax request
-		if (\Environment::get('isAjaxRequest'))
-		{
-			header('HTTP/1.1 204 No Content');
-			header('X-Ajax-Location: ' . $strLocation);
-		}
-		else
-		{
-			header('HTTP/1.1 303 See Other');
-			header('Location: ' . $strLocation);
-		}
-
-		exit;
+		static::redirect(\Environment::get('uri'));
 	}
 
 
@@ -1071,7 +1052,40 @@ abstract class Controller extends \System
 			header('Location: ' . $strLocation);
 		}
 
+		static::setCorsHeaders();
+
 		exit;
+	}
+
+	/**
+	 * Sends the correct HTTP headers for the back end rebuild search index feature.
+	 * To be removed as soon as we have a real web spider.
+	 */
+	public static function setCorsHeaders()
+	{
+		$strOrigin = \Environment::get('httpOrigin');
+
+		// Not a CORS request
+		if (!$strOrigin)
+		{
+			return;
+		}
+
+		$arrAllowedMethods = ['GET', 'HEAD', 'OPTIONS'];
+
+		// Protect and only allow requests from own known hosts and methods
+		if (in_array(\Environment::get('requestMethod'), $arrAllowedMethods))
+		{
+			$blnCheck = \Database::getInstance()->prepare('SELECT id FROM tl_page WHERE type=? AND dns=?')
+				->execute('root', preg_replace('@^https?://@', '', $strOrigin));
+
+			if ($blnCheck->numRows)
+			{
+				header('Access-Control-Allow-Headers: X-Requested-With');
+				header('Access-Control-Allow-Methods: ' . implode(', ', $arrAllowedMethods));
+				header('Access-Control-Allow-Origin: ' . $strOrigin);
+			}
+		}
 	}
 
 
--- a/system/modules/core/library/Contao/Controller.php
+++ b/system/modules/core/library/Contao/Controller.php
@@ -1058,8 +1058,7 @@ abstract class Controller extends \System
 	}
 
 	/**
-	 * Sends the correct HTTP headers for the back end rebuild search index feature.
-	 * To be removed as soon as we have a real web spider.
+	 * Send the correct HTTP headers when rebuilding the search index in the back end
 	 */
 	public static function setCorsHeaders()
 	{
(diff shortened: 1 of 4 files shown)


system/modules/core/library/Contao/Controller.php
```diff
@@ -1071,9 +1052,40 @@ public static function redirect($strLocation, $intStatus=303)
...
+			header('Access-Control-Allow-Headers: X-Requested-With, Contao-Rebuild-Search-Index');
+			header('Access-Control-Allow-Methods: GET, OPTIONS');
+			header('Access-Control-Allow-Origin: ' . \Environment::get('httpOrigin'));
+		}
+
+		// GET request (protect to only allow requests from own known hosts)
+		if (\Environment::get('requestMethod') == 'GET' && $strOrigin = \Environment::get('httpContaoRebuildSearchIndex'))
+		{
+			$blnCheck = \Database::getInstance()->prepare('SELECT id FROM tl_page WHERE type=? AND dns=?')
+				->execute('root', preg_replace('@^https?://@', '', $strOrigin));
```
leofeyer: This means that the back end URL **must** be entered in one of the root pages, doesn't it? However, it is not required to do so.

Toflar: It is required. Remember, this PR resolves **cross-domain** issues. You cannot have multiple domains in one installation without entering them in the root pages ;-)

fritzmg: Well you can leave out _one_ domain, technically. But it would not be advisable.

leofeyer:
So the entire validation is now based on the requesting domain being registered in a root page? Just asking because we were discussing sending tokens or cookies yesterday.

aschempp:
are you aware this is all not going to work if a page is cached?

Toflar:
Yes, because preflight requests can never contain authorization information.

> are you aware this is all not going to work if a page is cached?

Irrelevant because you're logged in in the back end when rebuilding the search index.

aschempp:
> Irrelevant because you're logged in in the back end when rebuilding the search index.

Not on the cross domain requests

aschempp:
Oh, and Varnish does not know about you being logged in…

Toflar:
Might need to move the code to the index.php then. Or use the modifyFrontendTemplate hook like @fritzmg did.

> Oh, and Varnish does not know about you being logged in…

Irrelevant. You need to configure the proxy correctly then.

Toflar:
Cached pages are now taken care of as well.

Toflar:
I just noticed that we also have CORS header issues when using the front end preview (logged in on `domain1` and previewing `domain2`). You cannot close the preview bar then because it's blocked. Shall I take care of this in this PR too?

fritzmg:
Looks good :). Minor nitpick: couldn't the `setCorsHeadersForCrossDomainSearchIndexRebuild` function simply be named `setCorsHeaders` so that it's less bulky? 😁 It's comment specifies that it is for cross domain search index rebuilding anyway.

leofeyer:
> You cannot close the preview bar then because it's blocked. Shall I take care of this in this PR too?

Yes please.

Toflar:
There are only two ways to fix the front end preview issue:

* Always send the `Access-Control-Allow-Origin` headers for domains of the same Contao installation
* Implement communication using `window.postMessage` which means we always have to include a script that handles FE preview on every page

I don't think that the second option really is an option because again, it would require to output a little js that is able to handle FE preview requests from cross domains on every single page and we cannot dynamically enable this obviously. It has to constantly be included on every single page.
If we go for the first option, however, we don't really have to build any complex support for the rebuild search index feature because by definition, all domains managed by the same Contao installation would always be allowed to request assets, scripts etc. cross domain wise.

leofeyer:
Since it turns out to be more complicated than expected, I think we should generate a separate PR for this.

Toflar:
But it affects this one as well. If we go for number 1 then only half of this PR is needed.

leofeyer:
Hm, you are right. My bad.

aschempp:
The correct solution for the preview bar is not to use iframes though, right?

Toflar:
Switched to `master`, pls reopen.

Toflar:
> The correct solution for the preview bar is not to use iframes though, right?

If possible, that would be best I guess, yes.

leofeyer:
Yes, it would be the best solution, but it is also the most complex one.

Toflar:
PR updated. Should be RTM now for 3.5.

ausi (approved):

aschempp (approved):

leofeyer (approved):

Toflar:
PR for the 10 billion times more beautiful and unit tested version for Contao 4: https://github.com/contao/core-bundle/pull/689

leofeyer:
```diff
               el.addClass('tl_green');
             },
             onFailure: function(xhr) {
+              var status = xhr.status;
+              if (status === 0) {
+                status = 'CORS request forbidden';
+              }
               el.addClass('tl_red');
-              el.set('text', '[' + xhr.status + '] ' + el.get('text'));
+              el.set('text', '[' + status + '] ' + el.get('text'));
             },
             url: el.getAttribute('data-url')
           }));
```

@Toflar We did not make these changes in Contao 4. Can I omit them?

Toflar:
I like them because they indicate what happened instead of just a red line not saying anything. Point is that the line will be colored red which to a regular user indicates that page might not exist. Then they open the url in the browser and it works as expected. If we add the hint `CORS request forbidden` then they have at least a hint and can google more on `CORS` :-)

leofeyer:
But "CORS request forbidden" should never occur after we have merged the changes, should it?

Toflar:
It does happen if the target page sends a redirect header.

leofeyer:
I see.

leofeyer:
I accidentally merged it into the wrong branch, so merged in 42515a6a7d08f130adf3a3a4dd86ce0f0e18c3c3 actually.

PR #8598, opened 2016-12-15
Contao 3.5.20

leofeyer:


discordier (approved):
LGTM :+1:

system/modules/core/library/Contao/Encryption.php
```diff
@@ -93,9 +98,14 @@ public static function encrypt($varValue, $strKey=null)
 	 * @param string $strKey   An optional encryption key
 	 *
 	 * @return string The decrypted value
+	 *
+	 * @deprecated Deprecated since Contao 3.5, to be removed in Contao 5.
+	 *             Use the DCA load and save callbacks to implement your own encryption routines.
 	 */
 	public static function decrypt($varValue, $strKey=null)
 	{
+		@trigger_error('Using Encryption::decrypt() has been deprecated and will no longer work in Contao 5.0. Use the DCA load and save callbacks to implement your own encryption routines.', E_USER_DEPRECATED);
```
aschempp: This message is misleading, because the methods are not related to the DCA

aschempp (changes requested):

system/modules/core/library/Contao/Encryption.php
leofeyer: The load and save callbacks are not related to the DCA?

discordier: He is talking about the `\Encryption` class in general and it's methods.

leofeyer: Ah, I see. So what do you suggest then?

discordier: Why not simply omitting the mentioning of DCa at all?
"Using Encryption::decrypt() has been deprecated and will no longer work in Contao 5.0. Implement your own encryption routines."

We could also write a chapter in the `UPGRADING.md` dedicated to the overall issue and explaining the reasons of the deprecation and simply pass this link into the message.

aschempp: I'm fine with @discordier's message. Though as mentioned in the respective ticket, people won't know where the message is coming from…

leofeyer:
All comments have been addressed. Please review again.

aschempp (approved):

ausi (approved):

system/modules/core/library/Contao/Encryption.php
```diff
@@ -47,9 +47,14 @@ class Encryption
 	 * @param string $strKey   An optional encryption key
 	 *
 	 * @return string The encrypted value
+	 *
+	 * @deprecated Deprecated since Contao 3.5, to be removed in Contao 5.
```
Toflar: This message feels really wrong to me. Please **NEVER** implement your own encryption routine but use existing and tested libraries for this :-) The message as it is now implies you should invent something on your own :-D

system/modules/core/library/Contao/Widget.php
```diff
@@ -1486,6 +1486,12 @@ public static function getAttributesFromDca($arrData, $strName, $varValue=null,
 			}
 		}
 
+		// Warn if someone uses the "encrypt" flag (see #8589)
+		if (isset($arrAttributes['encrypt']))
+		{
+			@trigger_error('Using the "encrypt" flag' . (!empty($strTable) && !empty($strField) ? ' on ' . $strTable . '.' . $strField : '') . ' has been deprecated and will no longer work in Contao 5.0. Use the load and save callbacks to implement your own encryption routine instead.', E_USER_DEPRECATED);
```
Toflar: Same here, please adjust to something like `Use well tested third party libraries such as phpseclib`.

Toflar (approved):

leofeyer:
The release has been scheduled for Monday, December 19th.

ausi (approved):

PR #8604, opened 2016-12-20
Fixed download element links did not consider generateFrontendUrl hook

Toflar:
I use the `generateFrontendUrl` hook to adjust the links but this is not considered in the download elements so I fixed this :-)

### DIFF
--- a/system/modules/core/elements/ContentDownload.php
+++ b/system/modules/core/elements/ContentDownload.php
@@ -87,7 +87,7 @@ class ContentDownload extends \ContentElement
 
 		global $objPage;
 		list(,$queryString) = explode('?', \Environment::get('request'), 2);
-		$strHref = $objPage->getAbsoluteUrl() . (($queryString != null) ? '?' . $queryString : '');
+		$strHref = $objPage->getAbsoluteUrl() . (($queryString !== null) ? '?' . $queryString : '');
 
 		// Remove an existing file parameter (see #5683)
 		if (preg_match('/(&(amp;)?|\?)file=/', $strHref))
--- a/system/modules/core/elements/ContentDownload.php
+++ b/system/modules/core/elements/ContentDownload.php
@@ -85,7 +85,9 @@ class ContentDownload extends \ContentElement
 			$this->linkTitle = specialchars($objFile->basename);
 		}
 
+		/** @var \PageModel $objPage */
 		global $objPage;
+
 		$strHref = $objPage->getAbsoluteUrl() . ((Environment::get('queryString') != '') ? '?' . Environment::get('queryString') : '');
 
 		// Remove an existing file parameter (see #5683)
--- a/system/modules/core/elements/ContentDownload.php
+++ b/system/modules/core/elements/ContentDownload.php
@@ -88,7 +88,7 @@ class ContentDownload extends \ContentElement
 		/** @var \PageModel $objPage */
 		global $objPage;
 
-		$strHref = $objPage->getFrontendUrl() . ((Environment::get('queryString') != '') ? '?' . Environment::get('queryString') : '');
+		$strHref = $objPage->getFrontendUrl() . (Environment::get('queryString') ? '?' . Environment::get('queryString') : '');
 
 		// Remove an existing file parameter (see #5683)
 		if (preg_match('/(&(amp;)?|\?)file=/', $strHref))
--- a/system/modules/core/elements/ContentDownload.php
+++ b/system/modules/core/elements/ContentDownload.php
@@ -88,7 +88,7 @@ class ContentDownload extends \ContentElement
 		/** @var \PageModel $objPage */
 		global $objPage;
 
-		$strHref = $objPage->getAbsoluteUrl() . ((Environment::get('queryString') != '') ? '?' . Environment::get('queryString') : '');
+		$strHref = $objPage->getFrontendUrl() . ((Environment::get('queryString') != '') ? '?' . Environment::get('queryString') : '');
 
 		// Remove an existing file parameter (see #5683)
 		if (preg_match('/(&(amp;)?|\?)file=/', $strHref))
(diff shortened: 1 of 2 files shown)


system/modules/core/elements/ContentDownload.php
```diff
@@ -85,7 +85,9 @@ protected function compile()
 			$this->linkTitle = specialchars($objFile->basename);
 		}
 
-		$strHref = \Environment::get('request');
+		global $objPage;
+		list(,$queryString) = explode('?', \Environment::get('request'), 2);
+		$strHref = $objPage->getAbsoluteUrl() . (($queryString != null) ? '?' . $queryString : '');
```
leofeyer: This should be `!== null`.

leofeyer (changes requested):

system/modules/core/elements/ContentDownload.php
leofeyer: Yes, we have always used strict type checking for `null`.

Toflar: Okay, fixed :)

leofeyer:
And wouldn't it be easier to use `Environment::get('queryString')` instead of splitting up the request URI?

Toflar:
Sure, fixed.

leofeyer (approved):

discordier (changes requested):
I do not like this approach.
You are now moving the dependency from the `Environment` class (representing the current request) to the `$objPage` (representing the bulk global document).
IMHO this will make the transition to fragments and the "no more superglobals" PR harder as the document object which ultimatively not available in all places.

Could you explain the original issue some more? I fail to see what you have adjusted in the `generateFrontendUrl`-HOOK to require this change.

Do you mean, that you altered the `absoluteUrl` of the `objPage` which the download element did not get because `\Environment::get('request')` returned something else?
If so, how were you able to access the current URL anyway?

Until this is discussed, I am against the change.

Toflar:
> You are now moving the dependency from the Environment class (representing the current request) to the $objPage (representing the bulk global document).

What the heck are you talking about? Let's try thinking about the goal and not the code before and after for a second: The goal ist to generate a link to the current page plus the `file` query parameter. Also, already existing parameters should be kept. In Symfony you would do:

```php
$parameters = $this->requestStack->getCurrentRequest()->query->all();
$parameters['file'] = $filePath; // Set or replace existing file query parameter
$url = $this->router->generate('route_name', $parameters);
```

So you would generate an URL **based on a given route**, not based on the current request URI string applying a set of regular expressions.

The very same is done in this PR now, so in my opinion it's nothing but correct :-)

discordier:
~I fail to see the difference, really...~

~Old (for moo.jpg):~
~1. Browser requests `https://example.com/foobar.html?bar=baz&file=img.jpg`~
~2. Contao fetches the current URI (relative in this case).~
~2. Contao changes the URI as `foobar.html?bar=baz&file=moo.jpg` by swapping the file parameter.~

~New (for moo.jpg):~
~1. Browser requests `https://example.com/foobar.html?bar=baz&file=img.jpg`~
~2. Contao fetches `$objPage` with id 1.~
~3.  `$objPage->getAbsoluteUrl()` generates `https://example.com/foobar.html` (absolute in this case).~
~4.  Appending current parameters generates `https://example.com/foobar.html?bar=baz&file=img.jpg`~
~5. Contao changes the URI as `https://example.com/foobar.html?bar=baz&file=moo.jpg` by swapping the file parameter.~

~So how the heck is this supposed to be generating a different result (aside from being now absolute)?
`$objPage->getAbsoluteUrl()` is supposed to always generate the current base url, isn't it?~

~The only thing where your change would make a difference is, when the DownloadElement is used with the wrong request in place (and therefore generates a link to a totally different URL), yet, in that case `$objPage` would also be the totally different object instance and therefore match up with the request again.~

~My main problem is, I don't see what problem got fixed here aside from now requiring a `$objPage` (which per definition is bound to the current request route) instead of the current "request route" by itself.~
~You are right, in Symfony I could generate the route as you wrote above.~
~Yet again, we do not have the symfony router to generate routes in Contao yet.~
~We have our own URL generator (used in `getAbsoluteUrl`, as you are using).~

All of above has been clarified by digging through the whole code and realizing that the (deprecated!) `generateFrontendUrl` HOOK is in the `PageModel::getAbsoluteUrl()` which was your initial problem.
This now puts things differently as you are now addung support to use a HOOK which will get removed in future versions and tying the dependency on the `objPage`.
Yet however, as the `ContentDownloads` is already bound on the page object, this is not that severe to me. It simply still feels wrong.

I have still some objections:
1. Your are not using the optional parameter parameter of `getAbsoluteUrl`, this should get adjusted?
2. Why are you now handling folders differently than files? See [system/modules/core/elements/ContentDownloads.php line 225](https://github.com/contao/core/pull/8604/files#diff-722911ba914d6930a67d9dd2051cb9fbR225) (You have to unfold the file below, sadly github is not autoexpanding).
If you really insist changing this behavior, then you must completely decouple the element from `Environment` and not only half of it. :smile_cat: 
3. You are now generating absolute URLs, where we were generating relative URLs before. I wonder if any problems might arise from there.

Toflar:
> Your are not using the optional parameter parameter of getAbsoluteUrl, this should get adjusted?

Why should I? I don't have any `$strParams`? I have a query string, that's not the same.



> Why are you now handling folders differently than files? See system/modules/core/elements/ContentDownloads.php line 225 (You have to unfold the file below, sadly github is not autoexpanding).
> If you really insist changing this behavior, then you must completely decouple the element from Environment and not only half of it. 😸

You're right, fixed in 9ebba0986bac4a9500c853b7aaf0be21cb12c730. Happens if the routine is copied instead of reused. No further comments on that.

> You are now generating absolute URLs, where we were generating relative URLs before. I wonder if any problems might arise from there.

Correct, thought that `\Environment::get('request')` was absolute too. Fixed in 9b60cbd02178a3755ba24eac6d52bbb78e1738de.

discordier (approved):

PR #8611, opened 2017-01-05
Contao 3.5.22

leofeyer:


discordier (approved):

Toflar (approved):

aschempp (changes requested):
The permission implementation is wrong. You're now adding the permissions to the user if adding it the the first group fails. But adding it to the user if the user does not have custom permissions does not make sense at all. We need to iterate over all groups to actually fix this.

system/modules/core/elements/ContentDownloads.php
```diff
@@ -146,7 +146,7 @@ protected function compile()
 					$arrMeta['title'] = specialchars($objFile->basename);
 				}
 
-				$strHref = \Environment::get('request');
+				$strHref = $objPage->getFrontendUrl() . (Environment::get('queryString') ? '?' . Environment::get('queryString') : '');
```
aschempp: The changelog entry for this seems missing?

system/modules/core/library/Contao/Model.php
```diff
@@ -682,7 +682,14 @@ public function refresh()
 	 */
 	public function detach($blnKeepClone=true)
 	{
-		\Model\Registry::getInstance()->unregister($this);
+		$registry = \Model\Registry::getInstance();
+
+		if (!$registry->isRegistered($this))
```
aschempp: The changelog entry for this seems missing as well

leofeyer:
> But adding it to the user if the user does not have custom permissions does not make sense at all.

True, this needs to be changed.

> We need to iterate over all groups to actually fix this.

This would be a change of behavior. We have been checking only the first group ever since, as we consider the first group to be the "main" group.

system/modules/core/elements/ContentDownloads.php
leofeyer: This is because the PR did not contain a change log entry. Also, there is not generally a change log entry for every commit, because that's what the commit history is for.

system/modules/core/library/Contao/Model.php
leofeyer: See above.

ausi (approved):

leofeyer:
Fixed the permission implementation in 439c08d08abd51c21631328a67f564599fc54831.

leofeyer:
The release has been scheduled for January 16th, 2017.

PR #8612, opened 2017-01-07
add logging for IP check

fritzmg:
The IP check of Contao can cause troubles for MacOS users, if their machine has both an IPv4 __and__ an IPv6 address and the host is also accessible via IPv4 __and__ IPv6. In this case MacOS (at least with FireFox, probably all browsers) establishes the HTTP connection randomly either via IPv4 or via IPv6.

So if a user logs in via IPv4 and then at any random time a request in the back end is made via IPv6, the user will be automatically logged out of Contao due to the IP mismatch. The system log will only contain the following entry:
```
User "…" has logged out
```
There will be no log entry at all as to why the user has suddenly been logged out. This PR would add a log entry for the IP check and thus make it at least more clear where the problem is. Log entries for the other cases (session ID check, hash check and timeout) could be added as well of course.

### DIFF
--- a/system/modules/core/library/Contao/User.php
+++ b/system/modules/core/library/Contao/User.php
@@ -260,8 +260,14 @@ abstract class User extends \System
 		$time = time();
 
 		// Validate the session
-		if ($objSession->sessionID != session_id() || (!\Config::get('disableIpCheck') && $objSession->ip != $this->strIp) || $objSession->hash != $this->strHash || ($objSession->tstamp + \Config::get('sessionTimeout')) < $time)
+		if ($objSession->sessionID != session_id() || $objSession->hash != $this->strHash || ($objSession->tstamp + \Config::get('sessionTimeout')) < $time)
 		{
+			return false;
+		}
+
+		if (!\Config::get('disableIpCheck') && $objSession->ip != $this->strIp)
+		{
+			$this->log('User authentication failed due to IP mismatch. Recorded IP: '.$objSession->ip.'. Current IP: '.$this->strIp, __METHOD__, TL_ERROR);
 			return false;
 		}
 
--- a/system/modules/core/library/Contao/User.php
+++ b/system/modules/core/library/Contao/User.php
@@ -267,7 +267,7 @@ abstract class User extends \System
 
 		if (!\Config::get('disableIpCheck') && $objSession->ip != $this->strIp)
 		{
-			$this->log('User authentication failed due to IP mismatch; recorded IP: '.$objSession->ip.', Current IP: '.$this->strIp, __METHOD__, TL_ERROR);
+			$this->log('User authentication failed due to IP mismatch; recorded IP: '.$objSession->ip.', current IP: '.$this->strIp, __METHOD__, TL_ERROR);
 			return false;
 		}
 
--- a/system/modules/core/library/Contao/User.php
+++ b/system/modules/core/library/Contao/User.php
@@ -267,7 +267,7 @@ abstract class User extends \System
 
 		if (!\Config::get('disableIpCheck') && $objSession->ip != $this->strIp)
 		{
-			$this->log('User authentication failed due to IP mismatch. Recorded IP: '.$objSession->ip.'. Current IP: '.$this->strIp, __METHOD__, TL_ERROR);
+			$this->log('User authentication failed due to IP mismatch; recorded IP: '.$objSession->ip.', Current IP: '.$this->strIp, __METHOD__, TL_ERROR);
 			return false;
 		}
 


discordier:
I will not go deeper into the problems with MacOS (as it is definitely behaving wrong, IPv6 should ALWAYS be preferred as per RFC) and AFAIR even MacOS and Windows follow the RFC on that one, so there seems to be a problem with your routing.
Logging the IP in login/logout actions however is a good move, not only for detecting your random logouts but also to pair up logins and logouts. We might as well want to log automatic logouts (due to ip violation) and expressive logouts (by clicking logout) differently.

fritzmg:
Windows and Linux definitely follow this. The problem only occurs with MacOS. Since it works with Linux and Windows from the same network and only MacOS shows this problem on that same network, I am inclined to rule out any routing issues. We observed this issue ourselves and it comes up every once in a while on the forum as well. Since it requires both the client and the host to have an IPv4 and an IPv6 address, it does not seem to happen that often yet (in the forum at least).

It appears this issues is only related to FireFox on MacOS after all, not to MacOS in general: https://bugzilla.mozilla.org/show_bug.cgi?id=920367

discordier:
See [here](https://labs.ripe.net/Members/emileaben/hampered-eyeballs) they state:
> On Lion, Chrome's "Happy Eyeballs" saves the day: the 300ms fallback-mechanism as described in the previous paragraph about Chrome just works. If the Firefox 7 "Happy Eyeballs" mechanism is activated, it works as well as Chrome. This is done by setting the value of `network.http.fast-fallback-to-IPv4` in `about:config` to true.

fritzmg:
So to solve the problem with Contao, you would need to set `network.http.fast-fallback-to-IPv4` to `false`, which is not ideal either.

discordier:
> So to solve the problem with Contao, you would need to set network.http.fast-fallback-to-IPv4 to false, which is not ideal either.

Nope, to **fix** the problem in **YOUR system** you would need to set this.
The **problem is not within Contao**. If you are coming from a different IP, you are another person, period.
You can still disable the IP check in system settings, however, that is not recommended (and in your case not even needed as the problem goes away by using either another browser or OS or configuring the problematic browser to not honor the OS decision).
To stress this out, **Contao IS NOT broken here but MacOS is**, there is nothing we can do at all aside from lowering security by not checking the client IP (which you already can do as stated above by changing the Contao config).

fritzmg:
> Nope, to fix the problem in YOUR system you would need to set this.

Yes, that's what I mean. Also it's not _my_ system, it's the system of MacOS users ;)

> The problem is not within Contao.

I know.

> You can still disable the IP check in system settings, however, that is not recommended

Exactly.

> (and in your case not even needed as the problem goes away by using either another browser or OS or configuring the problematic browser to not honor the OS decision)

I am not talking about me, I am talking about users that are affected by this in general.

> To stress this out, Contao IS NOT broken here but MacOS is, there is nothing we can do at all aside from lowering security by not checking the client IP (which you already can do as stated above by changing the Contao config).

You could track the IPv4 and IPv6 separately, but that would not make much sense imho.

discordier:
> I know. I never said there was a problem within Contao.

Sorry, then I have misinterpreted:
>> So to solve the problem with Contao

I agree that this PR makes sense as it provides more information. I further agree that it does not make sense to track the IPs differently. Let's conclude this by stating MacOS is broken and needs to get fixed and we provide better logging

xchs:
> Log entries for the other cases (session ID check, hash check and timeout) could be added as well of course.

👍 

We could also show the logout reason within the back end login screen, e.g.
```
Aus Sicherheitsgründen wurde Ihre Sitzung beendet. Bitte melden Sie sich erneut an.
```
```
For security reasons your session has expired automatically due to inactivity. Please log in again.
```

auto-logout

leofeyer:
> PS: Should I open a new ticket for this?

I don't think this is necessary. You can make the changes here.

aschempp:
Shouldn't this all get obsolete with Symfony Security?

discordier:
> Shouldn't this all get obsolete with Symfony Security?

You have to decide, in the end you and @bytehead are developing it. :smile_cat: 

I guess we can safely define adding of a log message as "bugfix" and push into a maintenance release.

fritzmg:
I am not quite sure how to do this actually. Since there is a redirect involved, I can`t just use
```php
\Message::add…
```
and be done with it. I'd have to store something in the session and then look for that in `BackendIndex::run`, though I am not sure if that would be the best way to do it.

leofeyer:
Since we are not logging IP addresses anymore, I guess we can close this issue.

PR #8613, opened 2017-01-09
[RFC] Backport contao/core-bundle#654

discordier:
This back ports the changes from contao/core-bundle/pull/654.

I really hope I got the ancient code style right. :smile_cat:

### DIFF
--- a/system/modules/core/library/Contao/System.php
+++ b/system/modules/core/library/Contao/System.php
@@ -692,23 +692,26 @@ abstract class System
 
 		// Use loadXML() instead of load() (see 7192)
 		$xml->loadXML(file_get_contents(TL_ROOT . '/' . $strName));
-		$return = "\n// $strName\n";
 
-		$objFileNodes     = $xml->getElementsByTagName('file');
-		$strCheckLanguage = strtolower($strLanguage);
+		$return       = "\n// $strName\n";
+		$objFileNodes = $xml->getElementsByTagName('file');
+		$strLanguage  = strtolower($strLanguage);
 
 		/** @var \DOMElement[] $objFileNodes */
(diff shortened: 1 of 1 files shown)


system/modules/core/library/Contao/System.php
```diff
@@ -692,9 +692,37 @@ protected static function convertXlfToPhp($strName, $strLanguage, $blnLoad=false
 
 		// Use loadXML() instead of load() (see 7192)
 		$xml->loadXML(file_get_contents(TL_ROOT . '/' . $strName));
-
 		$return = "\n// $strName\n";
-		$units = $xml->getElementsByTagName('trans-unit');
+
+		$objFileNodes     = $xml->getElementsByTagName('file');
+		$strCheckLanguage = strtolower($strLanguage);
```
ausi: Is there a reason why you didn’t reuse the `$strLanguage` here?
Like `$strLanguage = strtolower($strLanguage);`

ausi (approved):

system/modules/core/library/Contao/System.php
discordier: Not really, this is a left over from where I had everything in one method and had to preserve the original variable.
Can change it.

leofeyer: Also, we did not do a case-insensitive comparison before, did we?

ausi: The case-insensitive comparison is part of the contao/core-bundle#654 bugfix, because the XLIFF standard defines that languages are case-insensitive.

leofeyer (approved):

leofeyer:
Merged in 1cc5c5fb5e74636d5003ba846185c6bf711e60ec actually.

PR #8621, opened 2017-01-17
Contao 3.5.23

leofeyer:


Toflar (approved):

aschempp (commented):
Why are the changes from System and ImageSize not in the changelog?

system/modules/core/elements/ContentDownload.php
```diff
@@ -88,7 +88,14 @@ protected function compile()
 		/** @var \PageModel $objPage */
 		global $objPage;
 
-		$strHref = $objPage->getFrontendUrl() . (Environment::get('queryString') ? '?' . Environment::get('queryString') : '');
```
aschempp: I would actually go with something like this:

```php
$strHref = '#';
if ($objPage instanceof \PageModel)
{
	$strHref = $objPage->getFrontendUrl() . (Environment::get('queryString') ? '?' . Environment::get('queryString') : '');

	// Remove an existing file parameter (see #5683)
  	if (preg_match('/(&(amp;)?|\?)file=/', $strHref))
 	{
 		$strHref = preg_replace('/(&(amp;)?|\?)file=[^&]+/', '', $strHref);
 	}

	$strHref .= ((\Config::get('disableAlias') || strpos($strHref, '?') !== false) ? '&amp;' : '?') . 'file=' . \System::urlEncode($objFile->value);
}
```

aschempp (changes requested):

system/modules/core/elements/ContentDownload.php
leofeyer: Makes no difference at all IMO. Except now the link actually works in the BE and it does not in your suggested code.

PR #8647, opened 2017-02-17
Delete old search entry if new URL is more canonical

### LINKED ISSUE #8498: Duplicate entry for key 'url' INSERT INTO tl_search
Ähnlich dem #8460 .
Contao 3.5.17 hat zwei Seitenbäume, für Sprache de und en bei gleicher Domain.
Die Sprachkürzel in der URL sind nicht aktiviert! 
Home Alias bei "de" ist "index", bei "en" ist "home".

Ruft man nun mittels de und en Browser die Home Seiten ohne Alias auf, wird zweimal versucht die Url http://domain.xy/ einzutragen.

ausi:
Fixes #8498, #8460

### DIFF
--- a/system/docs/CHANGELOG.md
+++ b/system/docs/CHANGELOG.md
@@ -1,6 +1,13 @@
 Contao Open Source CMS changelog
 ================================
 
+Version 3.5.25 (2017-XX-XX)
+---------------------------
+
+### Fixed
+Delete an old search entry if the new URL is more canonical (see #8647).
+
+
 Version 3.5.24 (2017-01-19)
 ---------------------------
 
--- a/system/modules/core/library/Contao/Search.php
+++ b/system/modules/core/library/Contao/Search.php
@@ -186,17 +186,19 @@ class Search
 		{
 			if (strpos($arrSet['url'], '?') === false && strpos($objIndex->url, '?') !== false)
 			{
-				// the new URL is more canonical (no query string)
+				// The new URL is more canonical (no query string)
 				$objDatabase->prepare("DELETE FROM tl_search WHERE id=?")
 							->execute($objIndex->id);
 			}
 			elseif (substr_count($arrSet['url'], '/') > substr_count($objIndex->url, '/') || strpos($arrSet['url'], '?') !== false && strpos($objIndex->url, '?') === false || strlen($arrSet['url']) > strlen($objIndex->url))
 			{
-				$arrSet['url'] = $objIndex->url; // the current URL is more canonical (shorter and/or less fragments)
+				// The current URL is more canonical (shorter and/or less fragments)
+				$arrSet['url'] = $objIndex->url;
 			}
 			else
 			{
-				return false; // the same page has been indexed under a different URL already (see #8460)
+				// The same page has been indexed under a different URL already (see #8460)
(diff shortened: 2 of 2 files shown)


Toflar (approved):

leofeyer (approved):

aschempp (commented):
Instead of deleting and adding a new record, how about updating the URL so the existing entry is re-used?

ausi:
> Instead of deleting and adding a new record, how about updating the URL so the existing entry is re-used?

A new record is not always created, this depends on the `if ($objIndex->numRows)` on line 208. So I think deleting the record should be OK, otherwise the logic gets more complicated.

leofeyer:
That's not entirely true. We could run:

```php
$db
    ->prepare('UPDATE tl_search SET url=? WHERE ID=?')
    ->execute($arrSet['url'], $objIndex->id)
;
```

Then the query in line 201 would find the existing record.

leofeyer (changes requested):
If we really want to delete the existing record, we must also remove the child records from `tl_search_index`.

ausi:
> That's not entirely true. We could run:
> 
> ```php
> $db
>     ->prepare('UPDATE tl_search SET url=? WHERE id=?')
>     ->execute($arrSet['url'], $objIndex->id)
> ;
> ```
> 
> Then the query in line 201 would find the existing record.

I don’t think so. This would result in a `Query error: Duplicate entry '...' for key 'url' (UPDATE tl_search...` error. See the reproducing steps from https://github.com/contao/core/issues/8498#issuecomment-280591644

ausi:
> If we really want to delete the existing record, we must also remove the child records from `tl_search_index`.

Addressed in fb408700098cc185b9e14c5ef91311d45bf4df3a

leofeyer:
> I don’t think so. This would result in a `Query error: Duplicate entry '...' for key 'url' (UPDATE tl_search...` error.

You are right, of course.

leofeyer (approved):

aschempp:
Why would that result in a duplicate index issue when we update the URL?

ausi:
Because an entry with the new URL might already exist and the URL column has an unique index. I tried it out and got the `Duplicate entry '...' for key 'url'` error...

aschempp:
I don't understand that.
 1. we are searching for a database record by checksum
 2. if that record exists, we must delete it and not modify the URL, because the URL could exist
 3. then we create a record with that URL?

fritzmg:
@aschempp see https://github.com/codefog/contao-cookiebar/issues/21#issuecomment-263834643 for an example. The same page could be present multiple times in the `tl_search` table, with different GET parameters and different checksums. If you try to update the URL for one particular checksum, you might have duplicate URLs.

aschempp:
I do understand that, but how can this be an issue if we add a new database record with exactly that URL and checksum just a few code lines later?

ausi:
> I do understand that, but how can this be an issue if we add a new database record with exactly that URL and checksum just a few code lines later?

A few lines later there is a check if an entry with that URL already exists. Only if it doesn’t exist a new record is created, otherwise the existing record gets updated. And this record doesn’t have the same URL as the record we have deleted.

Issue #8643, opened 2017-02-09
Cache + insert_module|uncached + Custom template name = critical error


KaiserCh:
I've found an error in Contao 3.5.x in a rather odd situation.

If you use {{insert_module::*|uncached}} to insert a module (maybe displaying session-related infos) into a cached page while: 
- the page cache is enabled & filled
- the module uses a custom named template file (something like mod_newsreader_ext.html5)
- this template file is located in a subfolder of /templates

Contao will crash, the error message speaking of "template file not found".

If the template file uses the default file name but is located in a subfolder of /templates, it MIGHT load the default file from /system/modules/... Didn't test this. 
If the template file uses a custom file name and is located directly in /templates, everything works fine.

The source of the problem is the way template files are processed combined with Inserttag-replacing and the cache, starting at **Controller::getTemplate()**.
In this procedure, there is a call to the $objPage global variable. In an uncached environment, this variable is set and "knows" if the template might be located in a subfolder, since it knows the theme for the page, and the theme knows its theme template folder. But in a cached environment, $objPage is never set, so calling it here just returns something bogus. Maybe NULL, maybe a stdObject... but definitly nothing knowing anything about template paths. And since $objPage->templateGroup does not return anything, the TemplateLoader looks in the autoloader and /template/ for a file with the correct file name.

One possible solution would be to enforce the existence of $objPage inside Controller::getTemplate(). For any regular situation (uncached pages or cached paged without uncached insert_modules), nothing would happen since $objPage is (or should be) filled. For the buggy situation, the additional loading time & database queries for filling $objPage would be much better than a crash.

leofeyer:
This is odd indeed. @contao/developers

ausi:
Is `{{insert_module::*|uncached}}` a supported case for modules that need theme information?

I think the `uncached` flag should only be used on insert-tags that don’t need `$objPage`.

KaiserCh:
In the end, every module needs theme/template information, since every module or content element supports $this->customTpl per default. So either support |uncached for every module or for none.

Also: If I'm not mistaken, the issue also exists for content elements (and maybe more). I didn't test it, but since both Module and ContentElement use BaseTemplate::parse(), and both modules and CEs support custom templates from /template/subdirectory/, the results are predictable. **The bug should be triggerable with almost everything.** 
[...]

PR #8649, opened 2017-02-21
Fix insert_*::*|uncached by adding more data to cache file

KaiserCh:
Potential fix for #8643 
Added pageId and templateGroup to html cache file and using these in the controller to detect the otherwise missing data.

### DIFF
--- a/system/modules/core/classes/FrontendTemplate.php
+++ b/system/modules/core/classes/FrontendTemplate.php
@@ -250,13 +250,15 @@ class FrontendTemplate extends \Template
 			// Add the cache file header
 			$strHeader = sprintf
 			(
-				"<?php /* %s */ \$expire = %d; \$content = %s; \$type = %s; \$files = %s; \$assets = %s; ?>\n",
+				"<?php /* %s */ \$expire = %d; \$content = %s; \$type = %s; \$files = %s; \$assets = %s; \$objPageId = %s; \$objPageTemplateGroup = %s; ?>\n",
 				$strCacheKey,
 				(int) $intCache,
 				var_export($this->strContentType, true),
 				var_export($objPage->type, true),
 				var_export(TL_FILES_URL, true),
-				var_export(TL_ASSETS_URL, true)
+				var_export(TL_ASSETS_URL, true),
+				var_export($objPage->id, true),
+				var_export($objPage->templateGroup, true)
 			);
 
 			// Create the cache file
--- a/system/modules/core/controllers/FrontendIndex.php
+++ b/system/modules/core/controllers/FrontendIndex.php
@@ -435,6 +435,10 @@ class FrontendIndex extends \Frontend
 		$type = null;
 		$files = null;
 		$assets = null;
+		global $objPageId;
+		global $objPageTemplateGroup;
+		$objPageId = null;
+		$objPageTemplateGroup = "";
 
 		// Include the file
 		ob_start();
@@ -447,6 +451,7 @@ class FrontendIndex extends \Frontend
 
 			return;
 		}
+
 
 		// Define the static URL constants (see #7914)
 		define('TL_FILES_URL', $files);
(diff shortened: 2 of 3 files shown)


Toflar:
First of all, thank you for taking the time and trying to contribute. We need a lot more contributions 😃 
However, as I stated in the other issue, I don't think this is the right solution. I mean, I am very sure it's not. The way you implemented it already shows that it cannot be done easily. You register another global variable which is a deadly decision we cannot accept. However, I understand why you did this. It's because part of the logic is done in `PageRegular` and part of the logic in `FrontendIndex` which makes absolutely no sense from a software architectural point of view and which is why this needs to be fixed first. The elements that need the data must ensure they get it in both circumstances, when uncached and when cached. There might be a ton of other stuff other than the page and the template group that's needed for certain elements.
The thing is, we know how to solve it (ESI and isolated controllers) and we're already working on it in several pull requests (just search for "esi" on core-bundle). It just takes time.
Your code adds additional complexity, solves only a very tiny fraction of the problems and most importantly, would force us to port it to Contao 4 where we would need to rebuild some similar logic. Even if you consider this to be a bug fix, it is rather a feature than a bug fix to be honest. It adds new logic.

If you need to fix this issue for your Contao 3.5 project, you still can by building your own insert tag. But this is really something I cannot approve to be merged into the Contao core for outlined reasons above. The other core devs might have other opinions though.

ausi:
I agree with @Toflar, I don’t think it’s a good idea to merge this.

KaiserCh:
Well, it was just a quick fix for a rare bug. While rethinking the code, I've noticed a case, where my current solution would result in an error: if a cached page is deleted from page structure, but the page cache isn't cleared. \PageModel::findByPk would result in NULL and trying to add templateGroup to NULL would go boom.

I know it's ugly as hell, but at least it's fast. 
There are other ways to fix this, like using the objPageId from the cache file directly in \FrontendIndex::outputFromCache() to enforce the existence of $objPage. This would result in one additional DB query for every access to the cache. Not perfect, but still rather cheap. Getting the templateGroup would be much harder without a new global. There is no direct way to get it from the page object. The whole page (if it is of type PageRegular) needs to be parsed. Would be much easier if \PageRegular::getPageLayout() would be static or at least public. Then, it would be possible to create a new PageRegular-object and call ->getPageLayout($objPage).
Another way would be to violate DRI and recreate the $strCacheKey like \FrontendIndex::outputFromCache() does, re-open the cache file and use the objPageId & templateGroup directly, without setting them global.

Maybe it's really the best idea to simply ignore the problem.

aschempp:
> This would result in one additional DB query for every access to the cache. Not perfect, but still rather cheap.

What you probably didn't notice here is that the cache is currently build in a way that it does not require a database query at all. That means it does not hit the DB server, and it does not even establish a connection. It's not only about one query.

If I would go to fix this, just storing the templateGroup is not enough. But it would be possible to simply serialize the page object, similar as Symfony does with the token…

KaiserCh:
The DB connection is established as soon as (almost) any |uncached insert tag is triggered, but I didn't check, when or where the Database-static is first initialized. If I'm not mistaken, it will trigger as soon as a request to any \Model is made. So creating $objPage directly in outputFromCache() will should cost: 
- some file access, probably using the PHP opcode cache
- 1 initialized DB connection (if not set to permanent) 
- 1 query against a primary key, that should be cached by the RDBMS. 
Creating $objPage like I did in my pull request is just using the already initialized files and DB connection. 

Regarding the serialized page object: I thought about this, but for the caching bug it seems to be enough to know about the templateGroup. The whole page object would be one hell of a string. On the other hand, it would prevent the DB access. And @Toflar should be happy, because no additional global variables are needed. Just putting something like $objPageSerial = serialize($objPage) into the cached html file and accessing this data directly in outputFromCache() would work. 

Just some thoughts regarding the serialized data:
- Could this data contain anything sensible, that shouldn't be stored like this?
- How could bloating the cache html file impact performance?

leofeyer:
As discussed in Mumble on March 23rd, we consider this a known issue. We all agree that you are right, but the bugfix does not handle all cases and is very likely that it will cause further problems, therefore we prefer not to merge the PR.

PR #8654, opened 2017-02-25
[RFC] Use allowed_classes option in unserialize()

ausi:
Since PHP 7.0 there is a more secure way to use `unserialize()`, I think we should make use of it.

/cc @contao/developers

### DIFF
--- a/system/helper/functions.php
+++ b/system/helper/functions.php
@@ -327,15 +327,22 @@ function deserialize($varValue, $blnForceArray=false)
 		return $blnForceArray ? array() : '';
 	}
 
-	// Potentially including an object (see #6724)
-	if (preg_match('/[OoC]:\+?[0-9]+:"/', $varValue))
+	if (version_compare(PHP_VERSION, '7.0', '>='))
 	{
-		trigger_error('The deserialize() function does not allow serialized objects', E_USER_WARNING);
-
-		return $blnForceArray ? array($varValue) : $varValue;
+		$varUnserialized = @unserialize($varValue, array('allowed_classes' => false));
 	}
+	else
+	{
+		// Potentially including an object (see #6724)
+		if (preg_match('/[OoC]:\+?[0-9]+:"/', $varValue))
+		{
+			trigger_error('The deserialize() function does not allow serialized objects', E_USER_WARNING);
 
-	$varUnserialized = @unserialize($varValue);
+			return $blnForceArray ? array($varValue) : $varValue;
+		}
+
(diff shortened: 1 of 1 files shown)


Toflar:
Yeah, I was thinking about this but never made a PR😄 <x factor voice>It's a Yes for me!</x factor voice>

system/helper/functions.php
```diff
@@ -327,15 +327,22 @@ function deserialize($varValue, $blnForceArray=false)
...
 	}
 
-	// Potentially including an object (see #6724)
-	if (preg_match('/[OoC]:\+?[0-9]+:"/', $varValue))
+	if (version_compare(PHP_VERSION, '7.0', '>='))
 	{
-		trigger_error('The deserialize() function does not allow serialized objects', E_USER_WARNING);
-
-		return $blnForceArray ? array($varValue) : $varValue;
+		$varUnserialized = @unserialize($varValue, array('allowed_classes' => false));
```
leofeyer: This means that there will be no more warning, right?

ausi: Correct, objects will be instantiated as `__PHP_Incomplete_Class` objects.

leofeyer: So then we'll have a lot of `Warning: Class __PHP_Incomplete_Class has no unserializer in` messages?

aschempp:
You mean instead of an exception because classes are not supported? ;-)

system/helper/functions.php
ausi: I don’t think so. The unserialized result would just have `__PHP_Incomplete_Class` objects in it. My test didn’t show warning messages:

```php
unserialize('O:8:"stdClass":1:{s:3:"foo";s:3:"bar";}', ['allowed_classes' => false]);
/* Results in: 
__PHP_Incomplete_Class Object (
    [__PHP_Incomplete_Class_Name] => stdClass
    [foo] => bar
) 
*/
```

leofeyer: Hm, are we ok with not having a warning anymore?

ausi: Good question.
We could also keep the `preg_match` for PHP 7, and only add the `['allowed_classes' => false]` for extra security.

ausi:
Any reason why we didn’t merge this?

leofeyer:
It is tagged for Contao 4.5, so it will be merged.

leofeyer:
Merged in contao/core-bundle@26457736d416482a2e117a522645b0b9792de5c0.

Issue #8623, opened 2017-01-17
FormTextArea, specialchars & allowHtml


sebj54:
Hi,

I made a form on a website and I used a textarea with allowHtml set to to true. So I expected my field would save plain html in database instead of replacing `<` & `>` with `&lt;` & `&rt;`.

But this is not the case because of [this code](https://github.com/contao/core/blob/3.5.22/system/modules/core/forms/FormTextArea.php#L147) in `FormTextArea`:
```php
public function __get($strKey)
{
	switch ($strKey)
	{
		// [...]
		case 'value':
			return specialchars(str_replace('\n', "\n", $this->varValue));
			break;
		// [...]
	}
}
```

Is it the expected behavior?

Aybee:
I think this code is only for the content of the attribute "value".
```html
<textarea value="firts line\nsecond line<br>">
<textarea value="firts line
second line&lt;br&gt;">
```

sebj54:
That's right, it's used by Contao for the value attribute when generating a textarea widget. Personnaly, I use widgets to generate a form and to validate field values before saving them in database.
The problem with `FormTextArea` is that there is no other way to get its value without stripping tags.

Here are some extracts of my code:

```php
$strClass = $GLOBALS['TL_FFL'][$field['inputType']];
$objWidget = new $strClass($strClass::getAttributesFromDca($field, $name, $field['value'], '', '', null, true));

if ($isPost)
{
	$objWidget->validate();
}

$model->$widgetName = $objWidget->value;
$model->save();
```

leofeyer:
I think this is a bug introduced in 92a5488ddd90d9d1c15c2f7ee6f939abf7ae8a9f when fixing #7000. We should not generally apply `specialchars()` to the widget's value.

sebj54:
I can handle that issue and submit a pull request if you want!

leofeyer:
Thank you, but I don't think this will be necessary. It is a small fix and the ticket has already been assigned to the next milestone.

sebj54:
That's why I asked, thanks!

leofeyer:
Fixed in 3da97c6a942c2ca51c5b1dd0baf1f6cf44379019.

leofeyer:
[...]

PR #8670, opened 2017-03-16
Contao 3.5.25

leofeyer:


leofeyer:
The release has been scheduled for March 20th, 2017.

system/modules/core/forms/FormTextArea.php
```diff
@@ -144,7 +144,7 @@ public function __get($strKey)
 				break;
 
 			case 'value':
-				return specialchars(str_replace('\n', "\n", $this->varValue));
+				return str_replace('\n', "\n", $this->varValue);
```
ausi: In [form_textarea.html5:20](https://github.com/contao/core/blob/3e273a87f35ebb3ecafa36d695c3e380a1712b26/system/modules/core/templates/forms/form_textarea.html5#L20) the value gets printed without escaping `<?= $this->value ?>`, so this change introduces a security issue I think.

If you fill out a form with `</textarea>` in a textarea field and don’t fill out all required fields, you can break out of the textarea and inject XSS code.

ausi (changes requested):

system/modules/core/forms/FormTextArea.php
leofeyer: Fixed in 8b9a0175feddc2a7f9dcb3027bf1c46aa3bc3326.

ausi: What if someone uses a custom textarea template? I don’t think it’s a good idea to change this behavior in a bugfix release.

leofeyer: We are just restoring the previous behavior, because the last change was incorrect.

ausi: But isn’t this behavior already since 3.3.0?

leofeyer: You might want to read this: https://github.com/contao/core/issues/8623

ausi: I read it. But still, the behavior was since version 3.3.0 that `->value` of the textarea returned the encoded value. Every use of it that relies on that fact may break in an unsecure way.

leofeyer: You are right, the current behavior has been there since Contao 3.2.17. I have therefore reverted the changes in e603e0f4b567593ec39d53363c53cccd80ae7017.

ausi (approved):

PR #8695, opened 2017-04-08
fix cache key usage

### LINKED ISSUE #8694: inconsistent cache key generation
There is an inconsistency with the cache key generation and usage. When a page is generated for the page cache, `FrontendTemplate::addToCache` _always_ uses the __current host__ for the cache key: 
[system/modules/core/classes/FrontendTemplate.php#L216-L220](https://github.com/contao/core/blob/3.5.25/system/modules/core/classes/FrontendTemplate.php#L216-L220)
```php
$strCacheKey = \Environment::get('host') . '/empty.' . $objPage->language;
…
$strCacheKey = \Environment::get('host') . '/' . \Environment::get('request');
```
However, `Automator::generateConfigCache` _always_ uses either the _DNS setting_ from the page object or `*`: 
[system/modules/core/library/Contao/Automator.php#L539](https://github.com/contao/core/blob/3.5.25/system/modules/core/library/Contao/Automator.php#L539)
```php
$strBase = ($objPages->dns ?: '*');
```
So when the internal cache is active, the resulting `system/cache/config/mapping.php` will look like this for example:
```php
<?php

return array (
  '*/empty.fallback' => '*/empty.de',
  '*/empty.de' => '*/empty.de',
);
```
But the cache file was generated with `example.org/empty.de` as its cache key.

This then leads to the start page _never_ being served from the page cache, if the internal cache is built and no DNS setting was present in the website root.

fritzmg:
This fixes #8694.

Explanation:

* The `Automator` uses `$objPages->dns ?: '*'` [here](https://github.com/contao/core/blob/3.5.25/system/modules/core/library/Contao/Automator.php#L539) for creating the mappings of the _empty request_ cache key.
* However, `FrontendTemplate::addToCache` never actually uses `*` when creating the cache key (for the empty request), see [here](https://github.com/contao/core/blob/3.5.25/system/modules/core/classes/FrontendTemplate.php#L216). Thus the cache key for the empty request will always be `example.org/empty…`.
* Thus `FrontendIndex::outputFromCache` never actually finds a cache entry for the empty request [here](https://github.com/contao/core/blob/3.5.25/system/modules/core/controllers/FrontendIndex.php#L318-L431) since it will always be looking for `*/empty…` due to the mapping (if no DNS setting was made in the root page).

_Note:_ if the internal cache is __not__ in use or created, then [this Problem](https://github.com/contao/core/issues/7618#issuecomment-292555484) still remains.

### DIFF
--- a/system/modules/core/classes/FrontendTemplate.php
+++ b/system/modules/core/classes/FrontendTemplate.php
@@ -213,7 +213,7 @@ class FrontendTemplate extends \Template
 			// If the request string is empty, use a special cache tag which considers the page language
 			if (\Environment::get('request') == '' || \Environment::get('request') == 'index.php')
 			{
-				$strCacheKey = \Environment::get('host') . '/empty.' . $objPage->language;
+				$strCacheKey = ($objPage->domain ?: '*') . '/empty.' . $objPage->language;
 			}
 			else
 			{
@@ -251,7 +251,7 @@ class FrontendTemplate extends \Template
 			$strHeader = sprintf
 			(
 				"<?php /* %s */ \$expire = %d; \$content = %s; \$type = %s; \$files = %s; \$assets = %s; ?>\n",
-				$strCacheKey,
+				ltrim($strCacheKey, '*'),
 				(int) $intCache,
 				var_export($this->strContentType, true),
 				var_export($objPage->type, true),
--- a/system/modules/core/classes/FrontendTemplate.php
+++ b/system/modules/core/classes/FrontendTemplate.php
@@ -250,8 +250,8 @@ class FrontendTemplate extends \Template
 			// Add the cache file header
 			$strHeader = sprintf
 			(
-				"<?php /* %s */ \$expire = %d; \$content = %s; \$type = %s; \$files = %s; \$assets = %s; ?>\n",
-				ltrim($strCacheKey, '*'),
+				"<?php\n\n// %s\n\$expire = %d;\n\$content = %s;\n\$type = %s;\n\$files = %s;\n\$assets = %s;\n\n?>\n",
+				$strCacheKey,
 				(int) $intCache,
 				var_export($this->strContentType, true),
 				var_export($objPage->type, true),


system/modules/core/classes/FrontendTemplate.php
```diff
@@ -251,7 +251,7 @@ protected function addToCache()
 			$strHeader = sprintf
 			(
 				"<?php /* %s */ \$expire = %d; \$content = %s; \$type = %s; \$files = %s; \$assets = %s; ?>\n",
-				$strCacheKey,
+				ltrim($strCacheKey, '*'),
```
fritzmg: Just to comment on why this is necessary: when the cache key starts with `*`, the resulting PHP header of the cache file would look like this:
```php
<?php /* */empty.de */ $expire = 1494230344; $content = 'text/html'; $type = 'regular'; $files = ''; $assets = ''; ?>
```
This is invalid PHP syntax of course, thus I simply remove the `*` for the header.

system/modules/core/classes/FrontendTemplate.php
```diff
@@ -213,7 +213,7 @@ protected function addToCache()
 			// If the request string is empty, use a special cache tag which considers the page language
 			if (\Environment::get('request') == '' || \Environment::get('request') == 'index.php')
 			{
-				$strCacheKey = \Environment::get('host') . '/empty.' . $objPage->language;
+				$strCacheKey = ($objPage->domain ?: '*') . '/empty.' . $objPage->language;
```
leofeyer: Shouldn't this be `$objPage->dns`?

system/modules/core/classes/FrontendTemplate.php
leofeyer: How about `"<?php\n\n// %s\n\$expire = %d;\n\$content = %s;\n\$type = %s;\n\$files = %s;\n\$assets = %s;\n\n?>\n"` instead?

leofeyer (changes requested):

system/modules/core/classes/FrontendTemplate.php
fritzmg: No, `$objPage->dns` is only set for root pages. For other pages this information is saves in `->domain`. See [system/modules/core/models/PageModel.php#L868](https://github.com/contao/core/blob/3.5.25/system/modules/core/models/PageModel.php#L868)

fritzmg:
> How about "<?php\n\n// %s\n\$expire = %d;\n\$content = %s;\n\$type = %s;\n\$files = %s;\n\$assets = %s;\n\n?>\n" instead?

Sure, that should work too :)

system/modules/core/classes/FrontendTemplate.php
fritzmg: Sure, that should work too :). Should even work without the line breaks, I think.

fritzmg: I have changed it now.

leofeyer (approved):

leofeyer (approved):

PR #8701, opened 2017-04-19
Contao 3.5.26

leofeyer:


Toflar (approved):
Looks good to me.

system/modules/core/classes/FrontendTemplate.php
```diff
@@ -250,7 +250,7 @@ protected function addToCache()
 			// Add the cache file header
 			$strHeader = sprintf
 			(
-				"<?php /* %s */ \$expire = %d; \$content = %s; \$type = %s; \$files = %s; \$assets = %s; ?>\n",
+				"<?php\n\n// %s\n\$expire = %d;\n\$content = %s;\n\$type = %s;\n\$files = %s;\n\$assets = %s;\n\n?>\n",
```
ausi: Inserting the cache key as a PHP comment scares me a little bit because the URL is untrusted. There are ways to break out of single line comments like `?><?php`. I didn’t find a way to get these characters through the `Environment` class, but I’d prefer to play it safe and remove the comment completely from the cache file.

leofeyer: But then we cannot debug the cache key anymore. How about using `var_export()`?

ausi: Using `var_export()` together with `$cacheKey = %s;` should be OK I think.

system/modules/calendar/languages/en/default.xlf
```diff
@@ -11,6 +11,9 @@
       <trans-unit id="MSC.cal_next">
         <source>&amp;gt;</source>
       </trans-unit>
+      <trans-unit id="MSC.cal_timeSeparator">
+        <source>–</source>
```
aschempp: The separator previously contained a whitespace before and after the dash, is that intentionally no longer present?

system/modules/core/library/Contao/Environment.php
```diff
@@ -311,6 +311,11 @@ protected static function httpXForwardedHost()
 	 */
 	protected static function ssl()
 	{
+		if (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && in_array($_SERVER['REMOTE_ADDR'], trimsplit(',', \Config::get('proxyServerIps'))))
+		{
+		    return $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https';
```
aschempp: I still don't think this makes sense, because the remote IP being a proxy does not tell us the header is originating from the proxy. As stated in the issue (if I remember correctly), there's no way to know. The best would be for that specific case to use the `initconfig.php`. Anyway, if we can agree this is not unsafe, I'm fine with keeping it.

Toflar: It is secure and correct.

system/modules/core/classes/FrontendTemplate.php
leofeyer: Fixed in e8960dd36153d3ad4022c7bfca691b3e35760811.

system/modules/calendar/languages/en/default.xlf
leofeyer: Yes, see #8690.

aschempp (approved):

Issue #8742, opened 2017-07-11
"non-numeric value" bei Galerie mit PHP 7.1


Funbug:
Vermutlich ähnlich zu #8606

Ich habe den Fehler leider immer noch in einer 3.5.27 mit PHP 7.1:
Warning: A non-numeric value encountered in system/modules/core/elements/ContentGallery.php on line 304

Problem:
\Config::get('maxImageWidth') ist leer.

Mögliche Lösung (int):
$intMaxWidth = (TL_MODE == 'BE') ? floor((640 / $this->perRow)) : floor(((int)\Config::get('maxImageWidth') / $this->perRow));

leofeyer:
Gibt es `$GLOBALS['TL_CONFIG']['maxImageWidth']`-Einträge in Deiner `localconfig.php` oder `default.php`?

Funbug:
Ja, in der localconfig.php steht $GLOBALS['TL_CONFIG']['maxImageWidth'] = '';

leofeyer:
Dann weißt Du ja jetzt, woher der "non-numeric value" kommt.

Funbug:
Ich hab's da nicht reingeschrieben. Aber gut, danke. Problem gelöst. :)

fritzmg:
I can reproduce the issue of `$GLOBALS['TL_CONFIG']['maxImageWidth'] = '';` being written into the `localconfig.php`. Simply do the following:

1. Go to System » Settings » Files and images and set __Maximum front end width__ to a numeric value.
2. Save.
3. Remove the numeric value again (i.e. leave the input field empty).
4. Save.

However, I am not getting the `A non-numeric value` warning in the front end, when adding a content element with an image.

Funbug:
I only had the issue with galleries.

fritzmg:
You are right.

## Reproduction

1. Go to System » Settings » Files and images and set Maximum front end width to a numeric value, then save.
2. Remove the numeric value again (i.e. leave the input field empty), then save.
3. Add a gallery content element.
4. View the page with the gallery in the front end.

The following error will appear:

### Contao 3.5.27
```
Warning: A non-numeric value encountered in system\modules\core\elements\ContentGallery.php on line 304

#0 system\modules\core\elements\ContentGallery.php(304): __error(2, 'A non-numeric v...', 'C:\\xampp\\htdocs...', 304, Array)
#1 system\modules\core\elements\ContentElement.php(265): Contao\ContentGallery->compile()
[...]

PR #8743, opened 2017-07-11
make maxImageWidth mandatory

fritzmg:
Fix for #8742.


fritzmg:
Should we also change [this](https://github.com/contao/core/blob/3.5.27/system/modules/core/languages/en/tl_settings.xlf#L254-L256) to
```xml
<trans-unit id="tl_settings.maxImageWidth.1">
  <source>If the width of an image or movie exceeds this value, it will be adjusted automatically. Setting this to 0 disables the limit.</source>
</trans-unit>
```
or something to that effect?

leofeyer:
Yes.

fritzmg:
Added the translation change.

PR #8752, opened 2017-07-27
Make the file selector default to radio button

aschempp:
This will make the file tree behave like the page tree already does.
https://github.com/contao/core/blob/master/system/modules/core/widgets/PageSelector.php#L391


system/modules/core/widgets/FileSelector.php
```diff
@@ -347,6 +347,7 @@ protected function renderFiletree($path, $intMargin, $mount=false, $blnProtected
 						$return .= '<input type="checkbox" name="'.$this->strName.'[]" id="'.$this->strName.'_'.md5($currentFolder).'" class="tl_tree_checkbox" value="'.specialchars($currentFolder).'" onfocus="Backend.getScrollOffset()"'.$this->optionChecked($currentFolder, $this->varValue).'>';
 						break;
 
+					default:
 					case 'radio':
```
leofeyer: We don't need the `case 'radio':` anymore then.

leofeyer (changes requested):

system/modules/core/widgets/FileSelector.php
aschempp: `true`, but it's more explicit and the same in `PageSelector`…

Issue #8788, opened 2017-09-26
CSS combining not working with inline svg rgb definition


jacobro:
Contao 3.5.28. I use inline svg inside my css. I noticed, that after combining, contao adds a quote inside rgb color definition.
My css is:
```css
background-image: url("data:image/svg+xml;charset=utf8,<svg xmlns='http://www.w3.org/2000/svg'><path stroke='rgb(255, 255, 255)'></svg>");
```
after combining it looks as shown below
```css
background-image: url("data:image/svg+xml;charset=utf8,<svg xmlns='http://www.w3.org/2000/svg'><path stroke='rgb(255, 255, 255")'></svg>");
```
Combiner adds needless double quote just before rgb ending bracket

leofeyer:
What happens if you use single quotes for the image and double quotes for the attributes?

```css
background-image: url('data:image/svg+xml;charset=utf8,<svg xmlns="http://www.w3.org/2000/svg"><path stroke="rgb(255, 255, 255)"></svg>');
```

jacobro:
It outputs sth much more weird than before, as it still adds double quote before ending bracket and also changes first single quote after `url(` to double quote. I think combiner "thinks" that rgb ending bracket is the end of url definition.
```css
background-image: url("data:image/svg+xml;charset=utf8,<svg xmlns="http://www.w3.org/2000/svg"><path stroke="rgb(255, 255, 255")"></svg>');
```

leofeyer:
Fixed in a2be3cb372ae6761ac2deb3469157046ae2b4eb7.

PR #8790, opened 2017-09-27
[RTM] Improve url() handling in the Combiner

ausi:
See #8788

### DIFF
--- a/system/modules/core/library/Contao/Combiner.php
+++ b/system/modules/core/library/Contao/Combiner.php
@@ -383,10 +383,11 @@ class Combiner extends \System
 		$strDirname = dirname($arrFile['name']);
 		$strGlue = ($strDirname != '.') ? $strDirname . '/' : '';
 
-		return preg_replace_callback('/url\(("[^"\n]+"|\'[^\'\n]+\'|[^"\'\s()]+)\)/', function($matches) use ($strDirname, $strGlue)
+		return preg_replace_callback('/url\(("[^"\n]+"|\'[^\'\n]+\'|[^"\'\s()]+)\)/', function ($matches) use ($strDirname, $strGlue)
 		{
 			$strData = $matches[1];
-			if ($strData[0] === '"' || $strData[0] === '\'')
+
+			if ($strData[0] == '"' || $strData[0] == "'")
 			{
 				$strData = substr($strData, 1, -1);
 			}
@@ -417,7 +418,7 @@ class Combiner extends \System
 				$strData = '../../' . $glue . $strData;
 			}
 
-			$strQuote = $matches[1][0] === '\'' ? '\'' : '"';
+			$strQuote = $matches[1][0] == "'" ? "'" : '"';
 
 			return 'url(' . $strQuote . $strData . $strQuote . ')';
 		}, $content);
--- a/system/modules/core/library/Contao/Combiner.php
+++ b/system/modules/core/library/Contao/Combiner.php
@@ -418,7 +418,16 @@ class Combiner extends \System
 				$strData = '../../' . $glue . $strData;
 			}
 
-			$strQuote = $matches[1][0] == "'" ? "'" : '"';
+			$strQuote = '';
+
+			if ($matches[1][0] == "'" || $matches[1][0] == '"')
+			{
+				$strQuote = $matches[1][0];
+			}
+			elseif (preg_match('/[(),\s"\']/', $strData))
+			{
+				$strQuote = '"';
+			}
 
 			return 'url(' . $strQuote . $strData . $strQuote . ')';
 		}, $content);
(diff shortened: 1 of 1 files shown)


system/modules/core/library/Contao/Combiner.php
```diff
@@ -383,48 +383,43 @@ protected function fixPaths($content, $arrFile)
...
+					$strData = substr($strData, 3);
 				}
+
+				$glue = ($dir != '.') ? $dir . '/' : '';
+				$strData = '../../' . $glue . $strData;
 			}
 
-			$strBuffer .= 'url(' . $chunks[$i+1] . $strData . $chunks[$i+1] . ')';
-		}
+			$strQuote = $matches[1][0] === '\'' ? '\'' : '"';
```
leofeyer: This would always add double quotes even if the original string had no quotes at all.

leofeyer (changes requested):

system/modules/core/library/Contao/Combiner.php
ausi: Yes, that’s on purpose. We change the path so we should also add quotes (as we did in the past).

leofeyer: Did we really do this in the past? Because we really should not! Two unnecessary characters that have not been there before.

ausi: But they may be necessary, depending on the path we generate.

Yes, in the past we always used double quotes: https://github.com/contao/core/blob/3.5.28/system/modules/core/library/Contao/Combiner.php#L425

leofeyer: Yes, but that does no longer work with unencoded `data:` URLs, which is why I have changed it  (see #8788). We must use single quotes if the original used single quotes, double quotes if the original used double quotes and we can and should use no quotes if the original used no quotes.

ausi: If `data:` is used, the replacement gets completley skipped and nothing gets changed at all (in line 397).

If we replace `url(file.jpg)` with e.g. `url(../../files/some dir/file.jpg)` we would break it because whitespace (and some other characters) are not allowed in unquoted URLs.

leofeyer: Then we should probably check this? I don't know, its two unnecessary characters. 😄

ausi: I think it’s unnecessary to remove the quotes. For the gzipped size it could even be better if all `url()`s use quotes :)

leofeyer:
I did a quick performance analysis with 500 iterations.

OLD: 0.21314811706543
NEW: 0.22086501121521

So no performance concerns on my side.

system/modules/core/library/Contao/Combiner.php
leofeyer: This would do the trick:

```php
			if ($matches[1][0] != '"' && $matches[1][0] != "'" && !preg_match('/[()"\' ]/', $strDirname) && !preg_match('/[()"\' ]/', $strData))
			{
				$strQuote = '';
			}
			else
			{
				$strQuote = $matches[1][0] == "'" ? "'" : '"';
			}
```

leofeyer:
I really think we should try to omit the quotes if possible.

```php
$strQuote = $matches[1][0] == "'" ? "'" : '"';

// Omit the quotes if possible
if (!preg_match('/[(), "\']/', $strData))
{
	$strQuote = '';
}

return 'url(' . $strQuote . $strData . $strQuote . ')';
```

This will decrease the size of the COD style sheet from 49,653 to 49,613 bytes.

ausi:
Did you compare the size after `gzip` too?

I would use `[(),\s"\']` instead of `[(), "\']`

leofeyer:
> Did you compare the size after `gzip` too?

No. But if the source file is 40 bytes smaller, shouldn't the gzipped file be smaller, too?

> I would use `[(),\s"\']` instead of `[(), "\']`

Agreed.

ausi:
> But if the source file is 40 bytes smaller, shouldn't the gzipped file be smaller, too?

No. For example:

```sh
$ echo 'url("foo");url("bar");url("baz");url("foo");url("bar");url("baz");' | gzip | wc -c
44
$ echo 'url("foo");url("bar");url("baz");url(foo);url(bar);url(baz);' | gzip | wc -c
52
```

The source of the second example is smaller but the size after gzipping is larger.

leofeyer:
Only if there are mixed quotes. Mostly (I refer to the COD style sheet) all occurrences can be replaced, which results in:

```sh
echo 'url(foo);url(bar);url(baz);url(foo);url(bar);url(baz);' | gzip | wc -c
```

leofeyer:
Merged in 1dbdd894f61349d98bc5e692e5cfdc97b9691278. (I have also added unit tests.)

PR #8813, opened 2017-11-20
[RTM] Quote reserved words in database queries

ausi:
As discussed in contao/core-bundle#1106 and the Contao Mumble call, we want to escape all column names that are reserved keywords.

In Contao 4 we will use `\System::getContainer()->get('connection')->getDatabasePlatform()->quoteSingleIdentifier($name);`. In Contao 3.5 (this pull request) we only quote `rows`.

There is no need to fix the install tool in 3.5, as it uses quoted column names already.

@leofeyer Do you have an automated workflow to port this pull request to Contao 4.4, or should I do it by hand?

### DIFF
--- a/system/modules/core/library/Contao/Database.php
+++ b/system/modules/core/library/Contao/Database.php
@@ -396,7 +396,7 @@ abstract class Database
 	 */
 	public function isUniqueValue($strTable, $strField, $varValue, $intId=null)
 	{
-		$strQuery = "SELECT * FROM $strTable WHERE $strField=?";
+		$strQuery = "SELECT * FROM $strTable WHERE ".static::quoteColumnName($strField)."=?";
 
 		if ($intId !== null)
 		{
@@ -590,6 +590,37 @@ abstract class Database
 	public function getUuid()
 	{
 		return $this->get_uuid();
+	}
+
+
+	/**
+	 * Quote the column name if it a reserved word
+	 *
+	 * @param string $strName
+	 *
+	 * @return string
+	 */
+	public static function quoteColumnName($strName)
+	{
+		if (in_array(strtolower($strName), array('rows'), true))
+		{
+			$strName = '`'.$strName.'`';
+		}
+
+		return $strName;
+	}
+
+
+	/**
+	 * Quote a list of column names
+	 *
+	 * @param string[] $arrNames
+	 *
+	 * @return string[]
+	 */
+	public static function quoteColumnNames(array $arrNames)
+	{
+		return array_map(array(static::class, 'quoteColumnName'), $arrNames);
 	}
 
 
--- a/system/modules/core/library/Contao/Database.php
+++ b/system/modules/core/library/Contao/Database.php
@@ -396,7 +396,7 @@ abstract class Database
 	 */
 	public function isUniqueValue($strTable, $strField, $varValue, $intId=null)
 	{
-		$strQuery = "SELECT * FROM $strTable WHERE ".static::quoteColumnName($strField)."=?";
+		$strQuery = "SELECT * FROM $strTable WHERE ".static::quoteIdentifier($strField)."=?";
 
 		if ($intId !== null)
 		{
@@ -594,33 +594,20 @@ abstract class Database
 
 
 	/**
-	 * Quote the column name if it a reserved word
+	 * Quote an identifier if it is a reserved word
 	 *
 	 * @param string $strName
 	 *
 	 * @return string
 	 */
-	public static function quoteColumnName($strName)
+	public static function quoteIdentifier($strName)
 	{
-		if (in_array(strtolower($strName), array('rows'), true))
+		if ($strName == 'rows')
 		{
 			$strName = '`'.$strName.'`';
 		}
 
 		return $strName;
-	}
-
-
-	/**
-	 * Quote a list of column names
-	 *
-	 * @param string[] $arrNames
-	 *
-	 * @return string[]
-	 */
-	public static function quoteColumnNames(array $arrNames)
-	{
-		return array_map(array(static::class, 'quoteColumnName'), $arrNames);
 	}
 
 
--- a/system/modules/core/library/Contao/Database.php
+++ b/system/modules/core/library/Contao/Database.php
@@ -602,7 +602,7 @@ abstract class Database
 	 */
 	public static function quoteIdentifier($strName)
 	{
-		if ($strName == 'rows')
+		if (strtolower($strName) == 'rows')
 		{
 			$strName = '`'.$strName.'`';
 		}
--- a/system/modules/core/library/Contao/Database/Mysql.php
+++ b/system/modules/core/library/Contao/Database/Mysql.php
@@ -95,7 +95,7 @@ class Mysql extends \Database
 	 */
 	protected function find_in_set($strKey, $varSet, $blnIsField=false)
 	{
-		$strKey = \Database::quoteColumnName($strKey);
+		$strKey = \Database::quoteIdentifier($strKey);
 
 		if ($blnIsField)
 		{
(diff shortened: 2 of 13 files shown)


leofeyer:
I have a semi-automated way (creating split branches and then rebasing), so you don't have to do it by hand.

leofeyer:
I have a semi-automated way (creating split branches and then rebasing), so you don't have to do it by hand. But are the code changes actually the same in Contao 3 and Contao 4?

ausi:
> But are the code changes actually the same in Contao 3 and Contao 4?

Most of them are the same. The `quoteColumnName` method has to be different. Can you create a branch in the core-bundle where you merge these changes, and I add the additional modifications for Contao 4?

leofeyer:
Sadly this is not how it works. I can only create the split branches after the PR has been merged into Contao 3.5 and a new version has been tagged. So I guess we have to do it manually. 😢

ausi:
K, I do it :)

leofeyer:
Thank you very much.

leofeyer:
Before merging this we should check if this fix needs to be applied here as well (I think it does):

https://github.com/contao/core-bundle/commit/4de1d20ec55976549857067d79d589bf824a323b

ausi:
Yes, I think so. Needs to be applied in both *Mysql.php* and *Mysqli.php*.

leofeyer:
And this one as well: contao/core-bundle@541a8ca5a673bd561c49f5472f99287d34751438

leofeyer:
Unfortunately, there are even more problems. Sometimes the identifier is not a column but an expression such as `CONCAT(firstname, ' ', lastname)`, which leads to a "Column not found" exception.

Of course we could add something like this:

```php
// Not an identifier
if (!preg_match('/^[a-z0-9_$.]+$/', $strName))
{
	return $strName;
}
```

However this would leave the column names inside the `CONCAT()` function unquoted.

---

EDIT: The regex should be `A-Za-z0-9_$.` as mentioned below.

leofeyer:
@contao/developers I don't think that quoting identifiers is the correct solution as it leads to unforeseen issues. We have to discuss this topic again.

I am going to revert the changes in Contao 4.4.11 meanwhile.

ausi:
I still think quoting identifiers is the right solution and your fix with the regular expression is correct. Although we should allow uppercase letters too: `/^[a-z0-9_$.]+$/i`. Is there a specific reason why you included the `$` character?

> However this would leave the column names inside the `CONCAT()` function unquoted.

Which is correct IMO. If you as a developer pass such a string to the model you have to do the quoting yourself, because you are passing an SQL expression and not a column name.

leofeyer:
Apparently the dollar sign is a permitted character in an unquoted identifier:

https://dev.mysql.com/doc/refman/5.7/en/identifiers.html

DanielSchwiperich:
@ausi the PR also needs (if get recontinued) to take the creation into account:
https://github.com/madeyourday/contao-core/blob/e33738ab880063afc6453c17305479f58f98cf16/system/modules/core/drivers/DC_Table.php#L688

creating a new layout will otherwise not works because of the `rows` field in `tl_layout`
Same for  https://github.com/contao/core-bundle/pull/1262 if gets rereverted. 

There is also another issue regarding mariadb, will create an issue for it in core-bundle

ausi:
@DanielSchwiperich The set method is already handled, see https://github.com/contao/core/pull/8813/files#diff-a715adcad847fd730a703866083b4ae5R200

leofeyer:
As discussed in Mumble on January 18th, 2018, we want to merge this with the changes from contao/core-bundle@4de1d20 and contao/core-bundle@541a8ca.

ausi:
And add the regex check `/^[a-z0-9_$.]+$/i`

ausi:
I looked again and think we must not include the `.` so the correct regex is: `/^[a-z0-9_$]+$/i`

leofeyer:
What if it is `tl_content.singleSRC`?

ausi:
Same as with `CONCAT(...)`, the string `tl_content.singleSRC` is an expression and not a column name. The qouting would have to be done before. Like so `'tl_content.'.\Database::quoteColumnName('singleSRC');`

leofeyer:
Ok.

I wonder if the method should be called `quoteIdentifier()` instead of `quoteColumnName()`. The MySQL manual speaks of "unquoted identifiers" and the Doctrine method is called `quoteIdentifier()`, too. Any objections if I change it?

ausi:
> Any objections if I change it?

No, seems reasonable 👍
(but we should not forget to rename it in the Contao 4 pull request too)

system/modules/core/library/Contao/Database.php
```diff
@@ -594,15 +594,15 @@ public function getUuid()
...
 	 *
 	 * @param string $strName
 	 *
 	 * @return string
 	 */
-	public static function quoteColumnName($strName)
+	public static function quoteIdentifier($strName)
 	{
-		if (in_array(strtolower($strName), array('rows'), true))
+		if ($strName == 'rows')
```
ausi: @leofeyer why did you remove the `strtolower()`? 
This would make it no longer work for `Rows` or `ROWS`.

leofeyer:
```php
		// The identifier is quoted already
		if (strncmp($strName, '`', 1) === 0)
		{
			return $strName;
		}

		// Not an identifier
		if (!preg_match('/^[A-Za-z0-9_$]+$/', $strName))
		{
			return $strName;
		}
```

This is actually only required in Contao 4.4 as we are only quoting `rows` in Contao 3.5.

leofeyer (approved):

system/modules/core/library/Contao/Database/Mysql.php
```diff
@@ -95,7 +95,10 @@ protected function get_error()
 	 */
 	protected function find_in_set($strKey, $varSet, $blnIsField=false)
 	{
-		$strKey = \Database::quoteIdentifier($strKey);
+		if (preg_match('/^[A-Za-z0-9_$]+$/', $strKey))
```
ausi: This check is not necessary as it is handled by `quoteIdentifier()` itself.

system/modules/core/library/Contao/Database/Mysqli.php
```diff
@@ -85,7 +85,10 @@ protected function get_error()
 	 */
 	protected function find_in_set($strKey, $varSet, $blnIsField=false)
 	{
-		$strKey = \Database::quoteIdentifier($strKey);
+		if (preg_match('/^[A-Za-z0-9_$]+$/', $strKey))
```
ausi: Same here.

leofeyer:
Thank you @ausi.

system/modules/core/library/Contao/Database/Mysql.php
```diff
@@ -95,6 +95,8 @@ protected function get_error()
 	 */
 	protected function find_in_set($strKey, $varSet, $blnIsField=false)
 	{
+		$strKey = static::quoteIdentifier($strKey);
+
 		if ($blnIsField)
 		{
 			return "FIND_IN_SET(" . $strKey . ", " . $varSet . ")";
```
ausi, line 102: In this case we should probably use `"FIND_IN_SET(" . $strKey . ", " . static::quoteIdentifier($varSet) . ")";` because `$varSet` is then an identifier too.

system/modules/core/library/Contao/Database/Mysqli.php
```diff
@@ -85,6 +85,8 @@ protected function get_error()
 	 */
 	protected function find_in_set($strKey, $varSet, $blnIsField=false)
 	{
+		$strKey = static::quoteIdentifier($strKey);
+
 		if ($blnIsField)
 		{
 			return "FIND_IN_SET(" . $strKey . ", " . $varSet . ")";
```
ausi, line 92: Same here.

Issue #8814, opened 2017-11-21
{{picture::*}}-Inserttag not using width and height parameter


UB3RL33T:
Using the Picture-Tag I have noticed that, even if there are width and height parameters passed, the image will be displayed with it's natural width and height.

    {{picture::[imageOrUuid]?width=800&height=400&alt=sometext}}
Will ignore the passed width and height and use the images natural size.

Upon further investigation I have found that in 
    https://github.com/contao/core/blob/3e41ad8718f6dea9c10c8a9b0341ed86f2bbe2ae/system/modules/core/library/Contao/InsertTags.php#L1040
The Picture::create is being called with 0-values as $width and $height. 
This seems to be a bug as the documentation states, that width and height parameters can be passed to {{picture::*}}.

ausi:
The documentation is actually wrong. The `{{picture::*}}` insert tag only supports `size`, `alt`, `class`, `rel`, and `template`, see contao/core#7635.

`{{picture::*}}` is meant for responsive images, which needs a size ID instead of width/height. If you want to pass `width` and `height` you should use the `{{image::*}}` insert tag.

This should get fixed in the documentation.

PR #8815, opened 2017-11-21
Pass width and height to Picture::create

UB3RL33T:
See #8814


leofeyer:
@ausi /cc

ausi:
See https://github.com/contao/core/issues/8814#issuecomment-346155834

PR #8819, opened 2017-11-29
Do not encode chars from custom code field in exported style sheet

gmpf:
Preservation of HTML entered in the custom code field of a style sheet format definition was implemented here:
https://github.com/contao/core/pull/6667
https://github.com/contao/core/commit/c61c697278ca51f86fca96e08791073b45c197bc
However, the above only applies for the generated asset file. When the "Export style sheet" feature is used, all special characters in the created file will be converted to HTML entities.

To reproduce the issue, create a format definition (e. g. `.foo:before`) and add custom code (e. g. `content: "";`), then go back to the overview of style sheets and select "Export style sheet ID n". The resulting file will contain this:
```
.foo:before {
    content: &quot;&quot;;
}
```

To resolve the issue, check if the definition is not being compiled for export in addition to whether it is not being compiled for writing to a file, only then encode.


gmpf:
Related: https://github.com/contao/core-bundle/pull/1212

leofeyer:
Thank you @gmpf.

PR #8820, opened 2017-11-30
Fixed support for PHP 7.2 so that Argon2 works too

Toflar:
With the release of PHP 7.2, there's a new password hashing algorithm Argon2 available. Our `Encryption::test()` currently does not recognize that because it's built the wrong way around. It should not check for the ones it knows are password api compatible but check for the ones it knows are not. That ensures that we do not have to extend that check every time a new algorithm is introduced. The password api does that itself internally. 
I also removed all the legacy code which is handled way better by the `password-compat` polyfill so we know it behaves the same as does the core of php.

I also removed the bcrypt cost configuration. PHP itself defaults to a sane cost, I see no reason why we should not go with what the responsible crypto experts recommend.

This should also go to Contao 4.4 but without the polyfill as we require a php version >= 5.5 where the password api was introduced in php.

### DIFF
--- a/system/modules/core/library/Contao/Encryption.php
+++ b/system/modules/core/library/Contao/Encryption.php
@@ -201,10 +201,9 @@ class Encryption
 	 */
 	public static function test($strHash)
 	{
-		// We had SHA1, SHA256 and SHA512 before switching to password API compatible
-		// algorithms (first crypt() and later password_*()) so if it is was not one
-		// of the SHA algorithms, it is compatible.
-		return !(\in_array(\strlen($strHash), [40, 64, 128], true) && ctype_xdigit($strHash));
+		$info = password_get_info($strHash);
+
+		return 0 !== $info['algo'];
 	}
 
 
--- a/system/modules/core/library/Contao/Encryption.php
+++ b/system/modules/core/library/Contao/Encryption.php
@@ -185,61 +185,26 @@ class Encryption
 	 */
 	public static function hash($strPassword)
 	{
-		$intCost = \Config::get('bcryptCost') ?: 10;
+		@trigger_error('Using Encryption::hash() has been deprecated and will no longer work in Contao 5.0. Use password_hash() instead.', E_USER_DEPRECATED);
 
-		if ($intCost < 4 || $intCost > 31)
-		{
-			throw new \Exception("The bcrypt cost has to be between 4 and 31, $intCost given");
-		}
-
-		if (function_exists('password_hash'))
-		{
-			return password_hash($strPassword, PASSWORD_BCRYPT, array('cost'=>$intCost));
-		}
-		elseif (CRYPT_BLOWFISH == 1)
-		{
-			return crypt($strPassword, '$2y$' . sprintf('%02d', $intCost) . '$' . md5(uniqid(mt_rand(), true)) . '$');
-		}
-		elseif (CRYPT_SHA512 == 1)
-		{
-			return crypt($strPassword, '$6$' . md5(uniqid(mt_rand(), true)) . '$');
-		}
-		elseif (CRYPT_SHA256 == 1)
-		{
-			return crypt($strPassword, '$5$' . md5(uniqid(mt_rand(), true)) . '$');
-		}
-
-		throw new \Exception('None of the required crypt() algorithms is available');
+		return password_hash($strPassword, PASSWORD_DEFAULT);
 	}
 
 
 	/**
-	 * Test whether a password hash has been generated with crypt()
+	 * Test whether a password hash has been generated with a password API compatible
+	 * algorithm.
 	 *
 	 * @param string $strHash The password hash
 	 *
-	 * @return boolean True if the password hash has been generated with crypt()
+	 * @return boolean True if compatible
 	 */
 	public static function test($strHash)
 	{
-		if (strncmp($strHash, '$2y$', 4) === 0)
-		{
-			return true;
-		}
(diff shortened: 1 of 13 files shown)


Toflar:
TODO:
- [ ] Use `password_get_info()`
- [ ] Make sure the password fields are `varchar(255)`

Toflar:
Ping reviewers, PR updated.

system/modules/core/library/Contao/Encryption.php
```diff
@@ -185,61 +185,25 @@ protected static function initialize()
...
-			return true;
-		}
-		elseif (strncmp($strHash, '$5$', 3) === 0)
-		{
-			return true;
-		}
+		$info = password_get_info($strHash);
 
-		return false;
+		return 0 !== $info['algo'];
```
Toflar: There is no constant for unknown password algorithms :(

ausi (approved):

ausi:
Should we add a notice somewhere that `\Config::get('bcryptCost')` is no longer used?

Toflar:
> Should we add a notice somewhere that \Config::get('bcryptCost') is no longer used?

I thought about that but to be honest, I think 99.9% of all users never changed that value. And adding a deprecation notice would mean we still need to support it but only for bcrypt and not the other algorithms (and it has to be done everywhere I now use `password_hash()` directly) so I thought it's not worth it and over-complicates things.

ausi:
K, agreed 🙂

aschempp (approved):

system/modules/core/library/Contao/User.php
```diff
@@ -381,7 +381,7 @@ public function login()
 		// The password has been generated with crypt()
 		if (\Encryption::test($this->password))
```
leofeyer: In order to being able to deprecate the `Encryption::test()` method, we could use

```php
$info = password_get_info($this->password);

if ($info['algo'] > 0)
{
```

here. WDYT?

system/modules/core/library/Contao/Encryption.php
leofeyer: Why not `$info['algo'] > 0`?

Toflar: It's the same? I don't really care :D

system/modules/core/library/Contao/User.php
Toflar: Ah yes, wie could do that indeed. Then we can probably deprecate the whole `Encryption` class, can't we?

leofeyer: Nope, just the hashing part. Encrypting and decrypting remains untouched.

Toflar: Talking about Contao 4.

leofeyer: We do not (yet) have an alternative for encrypting and decrypting data in Contao 4.

Toflar: Of course we have, it's already documented? What are you talking about exactly?

leofeyer: So how do you encrypt and decrypt data in Contao 4?

```php
$string = 'Hello world';
$encrypted = Encryption::encrypt($string);
$decrypted = Encryption::decrypt($encrypted);
```

Can you point me to where we have documented how to do this without the `Encryption` library?

leofeyer: Ah, I see. The methods are deprecated without a concrete alternative. 😄

leofeyer: Then yes, we can deprecate the whole class.

Toflar: `Using Encryption::encrypt() has been deprecated and will no longer work in Contao 5.0. Use a third-party library such as OpenSSL or phpseclib instead.` is a pretty well described alternative 😄 

So everything ready here or what do you need?

leofeyer:
Implemented in Contao 4.4 in contao/core-bundle@815b38483af9244680d6f489a01a2952944685fd.

While porting the changes, I noticed that we are not using `password_needs_rehash()` yet. Since we are using `PASSWORD_DEFAULT`, it is likely that the algorithm will change in the future and then we should ask the users to re-enter their passwords, shouldn't we?

Toflar:
You're right, we should check and update the hash in those cases.

leofeyer:
Added in contao/core-bundle@1e2096d3a04adba0d146a3717f18e33cad8e6012.

leofeyer:
Backported in 7d1f0f1fc088acbcf3a44486fdea38e5b3e43891.

fritzmg:
This backport increased the minimum PHP version for Contao 3.5 from PHP 5.4.0 to PHP 5.5.0 because the [`password_verify`](http://php.net/manual/en/function.password-verify.php) function is only available in PHP >=5.5.0.

See also https://community.contao.org/de/showthread.php?69433-Nach-Update-Fatal-error-Call-to-undefined-function-Contao-password_verify()

Toflar:
We might add https://github.com/ircmaxell/password_compat.

leofeyer:
My bad, I forgot to update the `composer.json` accordingly. :cry:

fritzmg:
~~If the new PHP minimum version stays, then the README, docs.contao.org and the Contao Check need to be updated too. This should may be also be announced somewhere.~~

Ah you meant to add ircmaxell/password_compat :)

leofeyer:
The issue has been fixed in Contao 3.5.33 (see 60893ac5140a73770f54f2339d3279090eaa9497).

PR #8838, opened 2018-01-12
always refresh cron.txt

binfalse:
`cron.txt` contains the timestamp of the last cron run. The file is **only** updated with the current timestamp once cron is actually executed.
However, with Docker use cases you may throw away a container any time and start from a clean Contao instance (eg in case of updates, migrations etc), only mounting/linking persistent stuff into the container.
Therefore, the next cron run may only be in a few hours (database still knows `lastrun`), and the `cron.txt` is not created...
Thus, every browser request will always entail an ugly `404`, until the next cron run...

This patch will always write the `cron.txt` containing:

* the current timestamp, if cron is really executed
* the timestamp from the database, otherwise

Writing the `cron.txt` even if it already exists shouldn't be a big deal..?

See also Docker image at:
https://hub.docker.com/r/binfalse/contao/

### DIFF
--- a/system/modules/core/controllers/FrontendCron.php
+++ b/system/modules/core/controllers/FrontendCron.php
@@ -148,6 +148,12 @@ class FrontendCron extends \Frontend
 			$return = false;
 		}
 
+		// Otherwise make sure cron.txt contains the correct lastrun-time
+		else
+		{
+			$this->updateCronTxt($objCron->value);
+		}
+
 		$this->Database->unlockTables();
 
 		return $return;
--- a/system/modules/core/controllers/FrontendCron.php
+++ b/system/modules/core/controllers/FrontendCron.php
@@ -146,14 +146,13 @@ class FrontendCron extends \Frontend
 			$return = false;
 		}
 
-		// Otherwise make sure cron.txt contains the correct lastrun-time
+		// Make sure the cron.txt file contains the correct time (see #8838)
 		else
 		{
 			$time = $objCron->value;
 		}
 
 		$this->Database->unlockTables();
-
 		$this->updateCronTxt($time);
 
 		return $return;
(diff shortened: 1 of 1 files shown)


aschempp:
This will cause a file write operation on every access, but I think that's fine. But please move the line out of the if-else blocks then (just next to the `unlockTables`.

aschempp (approved):
This will cause a file write operation on every access, but I think that's fine. But we should move the line out of the if-else blocks then (just next to the unlockTables).

leofeyer:
> This will cause a file write operation on every access, but I think that's fine.

Well I don't.

binfalse:
The update on `cron.txt` will now be written after unlocking the cron table, not matter if it gets a new or an old value.

binfalse:
@leofeyer that means you're not convinced, yet?
Do you think that is too much overhead to write the `cron.txt`? Or too much io load?

I mean, it will only be written max once per minute - even on very busy contao instances with hundreds of cron jobs.. And dumping a simple number once per minute into a file shouldn't be too much work for a web server...? I mean for that very same request, the webserver typically writes ways more logs just for the request (and usually you have tons more requests before arriving at `cron.php`...)
Or are there any other concerns?

What do you think about reverting eacb91d again and instead keep the `else` from 754d366 and add a check if the file already exists? Something like:

```
// Otherwise make sure cron.txt contains the correct lastrun-time
else
{
	if (!file_exists ('system/cron/cron.txt'))
		$this->updateCronTxt($objCron->value);
}
```

However, with eacb91d the cron table won't be blocked while writing `cron.txt`

leofeyer:
As discussed in Mumble on January 18th, 2018, we want to merge this PR.

binfalse:
Cool. Let me know if I can do anything! :)

leofeyer:
Thank you @binfalse.

Issue #8848, opened 2018-02-07
generateSitemap does not properly handle port-components in URLs


binfalse:
When urls in Contao require a port (such as `https://example.com:1234`, eg. when running at an unusual port), the urls in the sitemap become invalid...
The Automator sends every single url through [`rawurlencode`](https://secure.php.net/manual/en/function.rawurlencode.php), see https://github.com/contao/core/blob/c7f0310ebd3f4e8b32a82f10f9ffa6827ab4b2a3/system/modules/core/library/Contao/Automator.php#L414 (properly to prevent invalid XML?)
But `rawurlencode` will convert much more than necessary for XML documents... Therefore a few entities are recovered afterwards, see https://github.com/contao/core/blob/c7f0310ebd3f4e8b32a82f10f9ffa6827ab4b2a3/system/modules/core/library/Contao/Automator.php#L415
However, colons are only recovered in combination with `://`... Thus, port components will become something like `https://example.com%3A1234`

Thus, the resulting sitemap contains lot's of invalid urls...

binfalse:
PR #8849 uses [`htmlspecialchars`](https://secure.php.net/manual/en/function.htmlspecialchars.php) instead of [`rawurlencode`](https://secure.php.net/manual/en/function.rawurlencode.php) to create XML-safe versions of the URLs, which should fix the problem.
(However, I'm not sure if anything else is relying on `rawurlencode`-like entities somewhere?)

aschempp:
Should we use CDATA instead of urlencode? Or is there a Google recommendation?

dmolineus:
The requirements for url  escaping are described here: https://www.sitemaps.org/protocol.html

fritzmg:
Only the path and query should be encoded probably, not the whole URI including the host and port.

leofeyer:
See #8849.

PR #8849, opened 2018-02-07
fix url-encoding in sitemap [see #8848]

binfalse:
to create the sitemap all contao-urls were `rawurlencode`d:
https://github.com/contao/core/blob/c7f0310ebd3f4e8b32a82f10f9ffa6827ab4b2a3/system/modules/core/library/Contao/Automator.php#L414

afterwards, some special entities were recovered (for example slashes and questionmarks):
https://github.com/contao/core/blob/c7f0310ebd3f4e8b32a82f10f9ffa6827ab4b2a3/system/modules/core/library/Contao/Automator.php#L415

however, colons were only recovered in combination with `://`...
thus, if the contao website is running at an unusual port (or behind a reverse proxy etc) all urls may include the port number, such as `:PORT`, which becomes `%3APORT` via `rawurlencode`, which in turn renders the sitemap-urls invalid...

i guess the urls need to be treated specially to obtain valid XML documents?

here i propose to skip the `rawurlencode` plus subsequent recovering of certain entities, and instead encode the urls XML-safe using:

    htmlspecialchars ($string, ENT_XML1, 'UTF-8');

see https://github.com/contao/core/issues/8848

### DIFF
--- a/system/modules/core/library/Contao/Automator.php
+++ b/system/modules/core/library/Contao/Automator.php
@@ -411,9 +411,7 @@ class Automator extends \System
 			// Add pages
 			foreach ($arrPages as $strUrl)
 			{
-				$strUrl = rawurlencode($strUrl);
-				$strUrl = str_replace(array('%2F', '%3F', '%3D', '%26', '%3A//'), array('/', '?', '=', '&', '://'), $strUrl);
-				$strUrl = ampersand($strUrl, true);
+				$strUrl = htmlspecialchars($strUrl, ENT_XML1, 'UTF-8');
 
 				$objFile->append('  <url><loc>' . $strUrl . '</loc></url>');
 			}


leofeyer:
@ausi /cc

ausi:
> i guess the urls need to be treated specially to obtain valid XML documents?

No, they need to be URL-encoded, see https://github.com/contao/core-bundle/pull/1095#issuecomment-332480369

ausi:
@binfalse can you please try if the following code works for your use case?

```php
$strUrl = explode('/', $strUrl, 4);
$strUrl[3] = rawurlencode($strUrl[3]);
$strUrl[3] = str_replace(array('%2F', '%3F', '%3D', '%26', '%25'), array('/', '?', '=', '&', '%'), $strUrl[3]);
$strUrl = implode('/', $strUrl);
$strUrl = ampersand($strUrl, true);
```

binfalse:
@ausi the snippet seems to be working fine (seems to produce the same result as `htmlspecialchars ($string, ENT_XML1, 'UTF-8');` for me...?)

leofeyer:
Here is the difference between our current code and `htmlspecialchars($string, ENT_XML1, 'UTF-8')`:

https://3v4l.org/9FfDi

@ausi Is it correct that `[]` do not need to be encoded? Then we could use `htmlspecialchars` here.

leofeyer:
Ok, so as briefly discussed with @ausi, using `htmlspecialchars()` is not a good solution, because it does not handle unicode characters. I have updated my test case accordingly:

https://3v4l.org/g7cZJ

So we should go for @ausi's solution (with an additional `isset($strUrl[3])` check in case there is neither path nor query and no trailing slash).

And we should probably decode `%5B` and `%5D` as `[` and `]`.

leofeyer:
Fixed in 36be049ac13fa4111c3da6ceba7bfdce2aee405b.

Issue #8798, opened 2017-10-11
Generating Folder Hashes on large file trees slows down Contao


funkjunkee:
https://github.com/contao/core/blob/2474ff9a8e0cf0f533818461d20930e49baaa9b0/system/modules/core/library/Contao/Folder.php#L372

We've observed through stacktrace monitoring (New Relic), that when a file is uploaded in the backend, then generating/updating the hash for the parent folder does take a very long time (several minutes per file) IF there's a lot of files scattered around in multiple folders witin the path of the file. According to our monitioring, this is due to the fact, that each file name found within the path (RecursiveIteratorIterator operation) is concattenated into a very long string, for which an md5 hash is generated. Going through the filesystem recursively is basically what's killing performance in our case.

I've tried to research if the \Folder::getHash function is used in any way to check a folders integrity, but wasn't able to find anything such, thus, I'm deceived to believe, that this function may not be of any good use after all. So, in case the hash is not used for checking the integrity of the filesystem, then wouldn't it make sense to skip recursive directory traversal and generate the hash ONLY for the path to the folder?

leofeyer:
> I've tried to research if the \Folder::getHash function is used in any way to check a folders integrity, but wasn't able to find anything such, thus, I'm deceived to believe, that this function may not be of any good use after all.

The hash is used to find folders moved e.g. via FTP when synchronizing the filesystem with the database.

> According to our monitioring, this is due to the fact, that each file name found within the path (RecursiveIteratorIterator operation) is concattenated into a very long string, for which an md5 hash is generated. Going through the filesystem recursively is basically what's killing performance in our case.

@contao/developers Can we optimize this? Maybe hash the file paths so the concatenated string does not become too long?

ausi:
Couldn’t the hash of a folder just be a hash of all children’s hashes combined?

Toflar:
Sounds like a blockchain 😄  Should work imho but might still take long when you have a lot of subfolders. But it should work if we only do it for the topmost level of each folder and start from the very bottom so hashes are updated bottom to top.

ausi:
like so?

- folder, hash(`abc` + `def` + `mno`) = `pqr`
  - file1, hash `abc`
  - file2, hash `def`
  - subfolder, hash(`ghi` + `jkl`) = `mno`
     - file1, hash `ghi`
     - file2, hash `jkl`

Toflar:
Exactly.

funkjunkee:
[...]

PR #8856, opened 2018-02-19
Improve performance of folder hashing

ausi:
As discussed in #8798

### DIFF
--- a/system/modules/core/library/Contao/Dbafs.php
+++ b/system/modules/core/library/Contao/Dbafs.php
@@ -183,7 +183,6 @@ class Dbafs
 				$objModel->type      = 'folder';
 				$objModel->path      = $objFolder->path;
 				$objModel->extension = '';
-				$objModel->hash      = $objFolder->hash;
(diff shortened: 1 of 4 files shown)


system/modules/core/library/Contao/Dbafs.php
```diff
@@ -708,6 +750,33 @@ public static function syncFiles()
...
+
+		$objChildren = \Database::getInstance()
+			->prepare("SELECT hash, name FROM tl_files WHERE path LIKE ? AND path NOT LIKE ? ORDER BY name")
+			->execute($strPath.'/%', $strPath.'/%/%');
+
+		if ($objChildren !== null)
+		{
+			while ($objChildren->next())
+			{
+				$arrHash[] = $objChildren->hash . $objChildren->name;
```
leofeyer, line 776: Any particular reason why you have not added a separator here?

ausi: The hash has always the same length, so there is no need for a seperator.

leofeyer:
Thank you @ausi. This is great!

PR #8858, opened 2018-02-20
Cast BE_PAGE_OFFSET cookie to an integer

ausi:
The `BE_PAGE_OFFSET` cookie should be cast to an integer.


leofeyer:
Thank you @ausi.

PR #8859, opened 2018-02-24
Fix bug : Call to a member function parse() on array

trcwebdesign:
Support array on objWidget


ausi:
How can `$objWidget` be an array? It gets created with `$objWidget = new $strClass($arrData);`.

Do you have a `loadFormField` or `validateFormField` hook that returns an array instead of an object?

trcwebdesign:
Yes. I am using isotope und the contao_debugger https://github.com/trcwebdesign/contao-debugger with the HOOKS which has led to this error. Therefore I have just catched it if it is an array or it will crash Contao.

ausi:
Returning an array for the form field hooks is not supported. You have to return a `\Widget` object.

PR #8884, opened 2018-05-30
add hook to alter comment form fields

backbone87:
ref https://github.com/contao/core-bundle/issues/1512

### DIFF
--- a/system/modules/comments/classes/Comments.php
+++ b/system/modules/comments/classes/Comments.php
@@ -254,6 +254,24 @@ class Comments extends \Frontend
 			'options'   => array(1=>$GLOBALS['TL_LANG']['MSC']['com_notify'])
 		);
 
+		// HOOK: add custom fields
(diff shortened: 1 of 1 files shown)


Toflar (approved):
CS changes need to be applied but otherwise this looks fine.

leofeyer:
Contao 3 is in the "security fixes only" phase of the long term support phases and since this is not a security fix, I am closing the ticket.

backbone87:
@leofeyer can you please read the related discussion in https://github.com/contao/core-bundle/issues/1512
without this hook its nearly impossible to make contao 3.5 comments to adhere to the DSGVO

fritzmg:
@backbone87 since Contao 3.5 won't get any updates anymore we might as well just instruct people to manually patch the `Comments.php` - since there is no danger that those changes would get lost. The patch could be distributed via the `contao-privacy` extension.

PR #8891, opened 2018-07-16
[minor] SCA

kalessil:
Micro-optimizations, and a case where array keys fragmentation can cause side-effects.


ausi:
This repository only gets security-relevant changes [since May 2018](https://contao.org/en/release-plan.html).

You should probably open the pull request in [contao/core-bundle](https://github.com/contao/core-bundle) instead.

kalessil:
My bad, thank you for the hint @ausi !

PR #8892, opened 2018-07-16
Add information that this repository is no longer active

ausi:
@contao/developers should we update the readme as well?


Toflar:
Ah nice! Yes I think the README should be updated too.

Toflar:
Also, remaining PR's should be cleaned up/moved. I don't think one can disallow PR's, can we?

bytehead (approved):

leofeyer:
You cannot disable PRs, only issues. Don't know if this makes sense though?

ausi:
I think it is better to keep issues enabled and adding the ISSUE_TEMPLATE.md from this pull request. Disabling issues could confuse people to think “Contao no longer acceppts issues”.

ausi:
I added a text for the readme too now.

xchs:
Should we also change the description of the repository?

contao_core

README.md
```diff
@@ -1,6 +1,9 @@
 Contao Open Source CMS
 ======================
 
+**This repository holds the source code for version 3.5 of Contao and doesn’t receive bug fixes anymore.**  
+**Please take a look at [contao/core-bundle][10] instead.**
```
leofeyer: I don't like this, because it is easy to overlook. We should replace the existing content instead and provide more information like in the updated CONTRIBUTORS.md file. WDYT?

ausi: Yes, makes sense :)

leofeyer: Will do.

leofeyer:
Thank you @ausi.

PR #8894, opened 2018-09-07
Hash the autologin token in the database

leofeyer:
This PR implements #8888.


ausi (approved):
👍

ausi:
Do we want to hash other tokens that are stored in the DB too?

leofeyer:
My first impulse was to hash all tokens, however, after re-reading the [stackoverflow post](https://stackoverflow.com/questions/549/the-definitive-guide-to-form-based-website-authentication/477578#477578), I do no longer thinks so. Only "password like" tokens need to be hashed. Also, unused opt-in tokens are purged after 24 hours now (GDPR), so hashing provides no additional value.

aschempp:
Would it be possible to migrate existing autologin tokens instead of completely breaking them?

leofeyer:
It would sure be possible, however, existing tokens are potentially compromised, therefore we should explicitly break them IMHO.

